PR #6, opened 2019-05-14
Updated Navbar

itsundef:


thestinger:
This is going to need more work. It can't have the hard-wired 17px margin to center the text and the padding around the text on the right needs to match the other links.

itsundef:
Removed hard-wired 17px && padding matches with other links.

thestinger:
@itsundef It looks very wrong on mobile. Try viewing the page in Chromium on Android. I don't think hard-wiring the size of the logo is going to work well and there might also be an issue with how flex is being used for this. It would also be good to fix the code style to match the existing CSS and HTML.

itsundef:
Fixed code style but ¯\_(ツ)_/¯ about the alternative to image resizing without media queries.

thestinger:
I really need it to scale properly based on the size rather than anything being hard-wired, like having it delicately configured to fit properly based on how the tested browsers are rendering it.

thestinger:
Basically, it needs to be set up to fill the available space, and be no bigger than that.

PR #7, opened 2019-05-17
Update index.html

spmedia:
- Add in Twitter card meta data
- Add in robots index/follow
- Add Website Schema

### DIFF
--- a/static/index.html
+++ b/static/index.html
@@ -1,6 +1,7 @@
 <!DOCTYPE html>
-<html lang="en" prefix="og: http://ogp.me/ns#">
+<html lang="en" prefix="og: http://ogp.me/ns#" itemscope itemtype="https://schema.org/WebSite">
     <head>
+        <meta name="robots" content="index, follow">
         <meta charset="utf-8"/>
         <title>GrapheneOS</title>
         <meta name="description" content="GrapheneOS is a security and privacy focused mobile OS with Android app compatibility."/>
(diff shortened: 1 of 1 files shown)


thestinger:
Twitter supports the Open Graph metadata when these aren't present so most of them aren't required unless it's overriding the Open Graph metadata. I don't think it's needed right now.

Setting `twitter:card` to `summary` also doesn't seem to be required since it's the default when the Open Graph metadata is present. It can be tested with https://cards-dev.twitter.com/validator.

I think `twitter:site` may be useful, but as far as I can tell they aren't currently using it and I couldn't find any information explaining what had happened with that. I'm curious if they removed it due to issues with phishing.

Is there a good reason to use the WebSite schema? I already dislike the duplication caused by the Open Graph metadata for the description even though it's currently the same. I'd rather not add it unless there's something using it that's not going to work with the existing metadata.

The robots directive should be redundant since it's already the default.

spmedia:
Schema and Twitter is just markup I always like to see sites have.

The twitter:card summary is from https://developer.twitter.com/en/docs/tweets/optimize-with-cards/overview/summary

I like to add a lot of meta data and schema to things. No worries if you dont want to implement any of it. It doesn't really 'hurt' anything per se by having it there. 

I looked at https://www.android.com/ and saw they were using Twitter card and WebSite schema so I figured it would be a good fit for Graphene as well.

thestinger:
I don't think I'll be implementing these because I think they're mostly redundant, but I'll keep an eye on `twitter:site` in case Twitter starts actually using that at some point again. It would be nice to have, but it doesn't do anything right now. If there's an example of anything using the metadata, I'll happily add it back, but I don't want to have anything without a clear purpose.

thestinger:
I also want to keep the metadata consistent across the pages, so if I do decide to add that, I'll do it for each page.

thestinger:
I went ahead with adding the twitter:site and twitter:creator metadata to all the pages based on evidence that crawlers use it despite it not being shown in the current generation Twitter cards. I feel like the rest of the metadata would be redundant.

spmedia:
Thanks Daniel!

PR #8, opened 2019-05-29
Add extra PGP optional key verification step

axelsimon:


thestinger:
This doesn't work with the current release of GPG using `--with-fingerprint` since it's not a command.

```
gpg: WARNING: no command supplied.  Trying to guess what you mean ...
pub   rsa4096 2012-12-06 [SC]
uid           Daniel Micay <danielmicay@gmail.com>
uid           Daniel Micay <daniel.micay@copperhead.co>
uid           Daniel Micay <security@attestation.app>
uid           Daniel Micay <security@seamlessupdate.app>
uid           Daniel Micay <security@grapheneos.org>
sub   rsa4096 2012-12-06 [E]
```

thestinger:
I need to think about this. The UX for this is really bad and the best option seems to be `gpg --import --import-options show-only thestinger.gpg` but people need to manually compare the fingerprint. GPG will also happily use any key in the keyring and I don't think there's a way to specify a specific one, rather than the user comparing the fingerprint afterwards.

thestinger:
@axelsimon The instructions now use signify and I added an out-of-band way of confirming the key:

https://grapheneos.org/install#obtaining-factory-images

It's a lot easier to do this now because it's tiny.

```
untrusted comment: GrapheneOS factory images public key
RWQZW9NItOuQYJ86EooQBxScfclrWiieJtAO9GpnfEjKbCO/3FriLGX3
```

It could be dropped down to `RWQZW9NItOuQYJ86EooQBxScfclrWiieJtAO9GpnfEjKbCO/3FriLGX3` too for verifying it out-of-band.

PR #9, opened 2019-05-29
Suggestion: instructions to use an alias to force recent fastboot

axelsimon:
A simple option when downloading the most recent Android Platform Tools to make sure to use the associated recent fastboot is to alias to it. These instructions (hopefully) make it simple for the reader to do so.


thestinger:
An alias won't work for the `flash-all.sh` script though. That only applies to the current interactive shell. You need to have it in the front of `PATH` for that.

thestinger:
It does work for the shell commands, but not `flash-all.sh`.

PR #14, opened 2019-07-03
Added the table of contents to the releases page

Peter-Easton:
Included the codenames this time around, as well as the release dates.


thestinger:
@Peter-Easton Can you make the same adjustment that I made in https://github.com/GrapheneOS/grapheneos.org/commit/6c68adf04358175e3a855d293f3c5e1b48693794? The nested sections should be nested inside the list item for the section.

PR #20, opened 2019-07-06
Added a stub entry for a FAQ page for later population as required.

Peter-Easton:


static/faq.html
```diff
@@ -0,0 +1,111 @@
...
+<!--Site footer below. Do not edit below this line.--> 
+        </div>
+        <footer>
+            <a href="/"><img src="https://grapheneos.org/logo.png" width="512" height="512" alt=""/>GrapheneOS</a>
+            <ul id="social">
+                <li><a href="https://twitter.com/GrapheneOS">Twitter</a></li>
+                <li><a href="https://github.com/GrapheneOS">GitHub</a></li>
+            </ul>
+        </footer>
+        <script type="application/ld+json">
```
thestinger: Can you leave out this application/ld+json part? It's only meant for the contact page.

Peter-Easton: Whoops, sure. Pushing it now.

static/faq.html
```diff
@@ -71,6 +71,7 @@ <h2 id="table-of-contents">
 <!--Content below -->
 
 <!--Site footer below. Do not edit below this line.--> 
+        </div>
```
thestinger, line 74: It seems you accidentally added a div here. Also, can you squash these into a single commit? Use `git rebase -i PARENT_COMMIT` (the one before your first commit) and then switch this 2nd commit from pick to squash. Do not reorder the lines, as that reorders commits.

Peter-Easton: Whoops. On it now.

PR #24, opened 2019-07-13
Added the verified boot section from the AndroidHardeningArchive to t…

Peter-Easton:
…he usage guide.


thestinger:
I don't want to add most of the this to the usage guide, as it's largely technical documentation rather than something aimed at end users. It's also somewhat out-of-date at this point.

The usage guide should be aimed at end users, not developers, and stick to explaining things to people without delving deep into technical details. They need to know the high-level security details and perhaps some reasoning behind why things are the way that they are but not all the technical details.

PR #25, opened 2019-07-13
Add a conditional to the signify command to only print "Verified" if …

Peter-Easton:
…indeed the factory.zip is indeed, valid in install.html


thestinger:
The `&&` is already a conditional and only runs the echo command if the exit status is zero.

PR #26, opened 2019-07-18
Add some of the safer questions to the FAQ to start with, avoiding co…

Peter-Easton:
…ntroversial questions & answers or borderline decisions.

### DIFF
--- a/static/faq.html
+++ b/static/faq.html
@@ -46,6 +46,26 @@
             <h2 id="table-of-contents">
                 <a href="#table-of-contents">Table of contents</a>
             </h2>
+                <ul>
+                    <li>
+                        <a href="#day-to-day-usage">Day To Day Usage</a>
+                            <ul>
+                                <li><a href="#is-it-ready">"Is it ready to be used as a daily driver OS?"</a></li>
+                            </ul>
+                    </li>
+                </ul>
+            
+            <h2 id="day-to-day-usage">
+                <a href="#day-to-day-usage">Day To Day Usage</a>
+            </h2>
+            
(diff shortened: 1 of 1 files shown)


thestinger:
@Peter-Easton It would be better to add these one-by-one because adjusting this much of it to my liking is too much and it will never get merged like this.

thestinger:
I'd also rather start with things not already covered by other documentation since the main use for this would be linking to it.

Peter-Easton:
Sure. I'll just start with the first one then.

static/faq.html
```diff
@@ -46,6 +46,26 @@ <h1 id="faq">
...
+            
+            <h2 id="day-to-day-usage">
+                <a href="#day-to-day-usage">Day To Day Usage</a>
+            </h2>
+            
+            <h3 id="is-it-ready">
+                <a href="#is-it-ready">"Is it ready to be used as a daily driver OS?"</a>
+            </h3>
+            <p>Resoundingly yes, Graphene OS is stable and feature-rich enough to be utilized as a daily 
+            driver. An app store like Yalp or F-Droid is not included, but can be added by downloading and
```
thestinger, line 66: The recommendation of Yalp and F-Droid really belongs in a separate answer.

Peter-Easton:
Sorry about the delay, 

Zenithium added a FAQ that uses and/or references strcat's writings on the reddit here on https://github.com/GrapheneOS/grapheneos.org/pull/29. The approach he's taken seems like a much safer way than the one I originally did when I sent the draft. I'm going to close this one.

PR #29, opened 2019-08-22
updated FAQ with Daniel's answers from the subreddit

Zenithium:
I felt that the community really needed a FAQ and I don't want Daniel to keep wasting his time responding to repetitive questions instead of doing more important work. Hopefully this should be of some use. The answers are all basically copy-pasted or paraphrased from Daniel's answers across the subreddit in an attempt to mitigate any misinformation from others and my own lack of knowledge on certain topics. I commited a day to going through all of the posts on the subreddit and picking out questions that I felt were frequently asked and/or important.

### DIFF
--- a/static/faq.html
+++ b/static/faq.html
@@ -46,6 +46,215 @@
             <h2 id="table-of-contents">
                 <a href="#table-of-contents">Table of contents</a>
             </h2>
+            <ol>
+            	<li><a href="#xda">Why isn't GrapheneOS on XDA?</a></li>
+            	<li><a href="#los">How is GrapheneOS different from LineageOS or other AOSP forks?</a></li>
+            	<li><a href="#funding">What will be the business model for GrapheneOS? How will the project fund development?</a></li>
+            	<li><a href="#stability">Can I use GrapheneOS as my daily driver? Is GrapheneOS stable?</a></li>
+            	<li><a href="#device-support">Will device X be supported?</a></li>
+            	<li><a href="#gapps">GApps?</a></li>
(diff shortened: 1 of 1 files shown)


Peter-Easton:
Once this is merged I can go back and fill in some of the others from my old draft, one at a time.

thestinger:
I started on an initial FAQ: https://grapheneos.org/faq. I'm going to close this pull request since the portion about device support is already part of the official FAQ and not all of it is relevant or up-to-date anymore.

Vanadium and the Vanadium WebView are separate apps, so users disabling Vanadium doesn't break anything as long as another browser is available to serve apps requiring the standard browser intents. In the future, the goal is to use Trichrome, which shares the code between them via a third app providing the shared libraries / resources.

I'd also prefer to have questions / answers submitted one-by-one so that I can more easily reject or accept the general idea behind it, provide a review of the contents and then edit it myself after merging something that's close to what I want.

Peter-Easton:
I still have that old draft I started some time ago. I've kept most of it and even moved a few things into their own section. Would you like me to start merging things one by one?

thestinger:
You could, but be aware that I'm not going to want all of it and may want to rewrite / change it substantially before merging it. It may take a lot of time to get it all merged.

thestinger:
You could work on a few at a time, and submit separate pull requests, so that it doesn't become blocked on a single one.

Peter-Easton:
Got it. Thank you!

If there's ones you don't need, don't like or don't want, I'm totally cool with that, I expected I would over-write with the initial drafts and then simply trim off what isn't necessary for the end.

PR #37, opened 2019-11-13
Initial overview added

warren-fisher:
Completed some initial work on adding an overview to the main page. Any recommendations or ideas appreciated. Looked through all posts on the Reddit and pointed out a few things _I_ thought should be addressed. If anything is wrong (I hope not), or unhelpful, or has been omitted please let me know.

### DIFF
--- a/static/index.html
+++ b/static/index.html
@@ -57,6 +57,31 @@
             href="https://attestation.app/about">about page</a> and <a
             href="https://attestation.app/tutorial">tutorial</a>. You can also extend these with
             support for your own builds.</p>
+            <h2 id="overview"><a href="#overview">Overview</a></h2>
+            <ul>
+                <li>
+                    GrapheneOS does <em>not</em> support Google play services. Some apps may not work, or only have partial functionality.
+                </li>
+                <li>
+                    GrapheneOS does <em>not</em> support rooting. This goes against its design goals and will <b>never</b> be supported.
+                </li>
+                <li>
+                    If you are looking to buy a phone to run GrapheneOS, the Pixel 3(xl) and 3a(xl) are the best options.
+                </li>
+                <li>
+                    GrapheneOS currently has no plans to support the Pixel 4 or Pixel 4XL
+                </li>
+                <li>
+                    This OS is 'barebones' and does not come with many preinstalled apps ontop of AOSP.
+                </li>
+                <li>
+                    You will most likely have problems running GrapheneOS on a carrier locked Pixel (eg. Verizon Pixels)
+                </li>
+                <!-- https://www.reddit.com/r/GrapheneOS/comments/bbm04f/has_anyone_gotten_sms_working_properly_with/ -->
+                <li>
+                    GrapheneOS works well using normal carriers, but does <strong>not</strong> support Google Fi and may have problems with Spring due to requiring a remote administration backdoor for it.
+                </li>
+            </ul>
             <h2 id="early-stage-of-development">
                 <a href="#early-stage-of-development">Early stage of development</a>
             </h2>


thestinger:
This is more of an FAQ and I don't really want to include that on the main page. Some is also already answered there.

warren-fisher:
I agree that it seems a lot like a FAQ and does not fit appropriately on the main page. One thing to note is that I cannot easily find any mentions of a lack of support for rooting devices so it may be helpful to add that somewhere on the page, perhaps with the device requirements. 

What did you have in mind for the overview? Perhaps more like bullet points of what will be mentioned further down on the page about "Early stage of development", "Roadmap" and "Device support"?

thestinger:
> One thing to note is that I cannot easily find any mentions of a lack of support for rooting devices so it may be helpful to add that somewhere on the page

This doesn't belong on the main page, and it's not clear what is even meant by that.

static/index.html
```diff
@@ -57,6 +57,31 @@ <h1 id="grapheneos">
...
+                    GrapheneOS does <em>not</em> support Google play services. Some apps may not work, or only have partial functionality.
+                </li>
+                <li>
+                    GrapheneOS does <em>not</em> support rooting. This goes against its design goals and will <b>never</b> be supported.
+                </li>
+                <li>
+                    If you are looking to buy a phone to run GrapheneOS, the Pixel 3(xl) and 3a(xl) are the best options.
+                </li>
+                <li>
+                    GrapheneOS currently has no plans to support the Pixel 4 or Pixel 4XL
```
thestinger, line 72: This isn't really what was stated.

static/index.html
```diff
@@ -57,6 +57,31 @@ <h1 id="grapheneos">
...
+            <h2 id="overview"><a href="#overview">Overview</a></h2>
+            <ul>
+                <li>
+                    GrapheneOS does <em>not</em> support Google play services. Some apps may not work, or only have partial functionality.
+                </li>
+                <li>
+                    GrapheneOS does <em>not</em> support rooting. This goes against its design goals and will <b>never</b> be supported.
+                </li>
+                <li>
+                    If you are looking to buy a phone to run GrapheneOS, the Pixel 3(xl) and 3a(xl) are the best options.
```
thestinger, line 69: This is already stated below.

static/index.html
```diff
@@ -57,6 +57,31 @@ <h1 id="grapheneos">
...
+                </li>
+                <li>
+                    GrapheneOS currently has no plans to support the Pixel 4 or Pixel 4XL
+                </li>
+                <li>
+                    This OS is 'barebones' and does not come with many preinstalled apps ontop of AOSP.
```
thestinger, line 75: It's already stated that the OS is in an early stage of development and I don't want to state things like this as if it's part of the goals rather than a consequence of that.

static/index.html
```diff
@@ -57,6 +57,31 @@ <h1 id="grapheneos">
...
+                </li>
+                <li>
+                    You will most likely have problems running GrapheneOS on a carrier locked Pixel (eg. Verizon Pixels)
```
thestinger, line 78: This isn't clear and also doesn't belong on the main page. It's not an attribute of the OS.

static/index.html
```diff
@@ -57,6 +57,31 @@ <h1 id="grapheneos">
...
+                </li>
+                <!-- https://www.reddit.com/r/GrapheneOS/comments/bbm04f/has_anyone_gotten_sms_working_properly_with/ -->
+                <li>
+                    GrapheneOS works well using normal carriers, but does <strong>not</strong> support Google Fi and may have problems with Spring due to requiring a remote administration backdoor for it.
```
thestinger, line 82: Similar to the previous one. It doesn't belong on the main page but rather somewhere else more specific documenting carrier support.

static/index.html
```diff
@@ -57,6 +57,31 @@ <h1 id="grapheneos">
             href="https://attestation.app/about">about page</a> and <a
             href="https://attestation.app/tutorial">tutorial</a>. You can also extend these with
             support for your own builds.</p>
+            <h2 id="overview"><a href="#overview">Overview</a></h2>
+            <ul>
+                <li>
+                    GrapheneOS does <em>not</em> support Google play services. Some apps may not work, or only have partial functionality.
```
thestinger, line 63: This is not how I want to explain something like this and it's better done in a usage guide section about it.

static/index.html
```diff
@@ -57,6 +57,31 @@ <h1 id="grapheneos">
...
+                </li>
+                <li>
+                    GrapheneOS does <em>not</em> support rooting. This goes against its design goals and will <b>never</b> be supported.
```
thestinger, line 66: Not clear what this means and it definitely doesn't belong on the main page. It also has to be properly explained if it's written somewhere, but I'm not sure it belongs on the web site at all.

thestinger:
By overview I mean a better description of what the project is focused on and what it does. In general, I want people to take the section stating that the OS is in an early stage of development more seriously and lower their expectations about documentation and the state of the OS for the time being. It doesn't have a proper development team or infrastructure yet and the focus is not trying to promote it.

warren-fisher:
Thanks for taking your time to read through my pull request - even if it was not up to standards it gave me some ideas for why it was not appropriate for this website. 

In particular to the "lack of support for rooting devices" I meant that I often see people asking on the subreddit if they can run Xposed, AFWall+, Magisk, etc... and associated "features" that require rooting. I merely meant that it could be helpful to point this out somewhere on the main page since it seems to be a common misconception with the OS. If it does not belong on the main page, where would it belong? Perhaps if there was a FAQ page this could solve this issue and other common misconceptions; although it does seem like most of those questions get answered fairly rapidly on the subreddit so it may not be necessary.

static/index.html
warren-fisher: I agree with this sentiment, the way I phrased it made it seem like a "downside" of the project.

PR #43, opened 2020-02-14
Add the udev rules to the installation prerequisites.

Peter-Easton:


thestinger:
The `adbusers` group isn't required to permit using devices from local sessions, so you shouldn't need to include it. It's only needed to permit using devices that are plugged in locally from a remote session. That's generally how udev permissions work. It already grants permission to access all the local devices for local sessions.

Peter-Easton:
Got it. I'll erase that part...

thestinger:
@Peter-Easton It looks like the `adbusers` instructions are still included.

Peter-Easton:
Whoa, whoops, sorry about that. Should be fixed now.

static/install.html
```diff
@@ -82,6 +83,31 @@ <h2 id="prerequisites">
...
+            <p>Linux distributions that use udev for device management, such as Fedora, Ubuntu,
+            and Debian do not include and require correct udev rules in order to permit the
+            software on the computer to talk to the device if running adb or fastboot as a non-
+            root user. Running fastboot as non-root without the prerequisite udev rules installed
+            will result in fastboot hanging with a <code>waiting for device</code> error message
+            even with the device plugged in and seemingly properly configured.</p>
+
+            <p>The easiest way to obtain and add the correct udev rules to your computer is to
+            run the following commands to download the android rules, move them to the udev rules
+            directory and reload the rules, then add the <code>adbusers</code> usergroup and
```
thestinger: Need to remove this bit too.

thestinger (changes requested):

static/install.html
Peter-Easton: Whoops. I removed the part about the adbusers usergroup but left the intro paragraph and the part about Fedora and Debian.

thestinger:
This is now covered for the supported operating systems.

PR #44, opened 2020-02-15
Add Q&A: securing data at rest, and FBE vs FDE

Peter-Easton:
Short little part about the encryption. This part is a little lengthy so I completely understand if it's not desired or it might do better in a separate page.

### DIFF
--- a/static/faq.html
+++ b/static/faq.html
@@ -602,6 +602,14 @@
             two-word passphrase from a wordlist of 2048 words at the same rate would take close to
             37,000,000,000 or <em>thirty seven thousand million</em> years.</p>
 
+            <p>A nine-point pattern unlock grid with between four and six dots allows for 34,792
+            possibilities, which in theory would require over twelve million years to systematically
+            brute-force at the rate of one attempt per day. Unfortunately, it's worth noting that
+            in most cases forcing a human being to bear the responsibility of generating their own
+            lockscreen patterns usually results in the user creating familiar or predictable
+            patterns which often are easy for a computer to guess, so lock patterns are often far
+            from ideal in practice.</p>
+
             <p>GrapheneOS allows passwords of up to 64 characters to be used.</p>
 
             <h3 id="fbe-over-fde">
--- a/static/faq.html
+++ b/static/faq.html
@@ -88,6 +88,9 @@
                         <li><a href="#firewall">Does GrapheneOS provide a firewall?</a></li>
                         <li><a href="#ad-blocking">How can I set up system-wide ad-blocking?</a></li>
                         <li><a href="#ad-blocking-apps">Are ad-blocking apps supported?</a></li>
+                        <li><a href="#securing-data-at-rest">How does GrapheneOS secure data at
+                        rest?</a></li>
+                        <li><a href="#fbe-over-fde">Why does GrapheneOS use FBE over FDE?</a></li>
                     </ul>
                 </li>
                 <li>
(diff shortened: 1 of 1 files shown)


static/faq.html
```diff
@@ -291,6 +294,55 @@ <h2 id="default-connections">
                 </li>
             </ul>
 
+            <h2 id="securing-data-at-rest">
+                <a href="#securing-data-at-rest">How does GrapheneOS secure data at rest?</a>
+            </h2>
+            <p>GrapheneOS makes full use of the hardware-bound encryption in modern Android devices.
+            The Titan M, a specially designed hardware security module on Pixel 3 handsets handles
```
inthewaves: "HSM" is used below without any introduction, so we should introduce it here. Also, the Titan M is present on the 3a
```suggestion
            The Titan M, a specially designed hardware security module (HSM) on Pixel 3 and 3a handsets, handles
```

static/faq.html
```diff
@@ -291,6 +294,55 @@ <h2 id="default-connections">
...
+            about the size of a grain of rice and is difficult to attack. Each file is encrypted
+            with AES-256-XTS, and the filesystem metadata is encrypted with AES-256-HEH.</p>
+
+            <p>File Based Encryption on GrapheneOS is always on, enforced, and cannot be switched
+            off or fall back to plaintext. The Titan M hardware security module has its own internal
+            timer which cannot be accessed from the host operating system, and can enforce
+            progressively lengthening delays between guesses at the password. On enough failed
+            guesses, the Titan M can increase the time between guesses to allow only one guess
+            per day. At this rate, to search through all the combinations of a four-digit pin code
+            from 0000 to 9950 would take over 27 years, or for a relatively weak two-word passphrase
```
inthewaves: Why 9950 and not 9999?

static/faq.html
```diff
@@ -291,6 +294,55 @@ <h2 id="default-connections">
...
+            it offline, and instead must go through the specific HSM on the phone. The HSM is
+            about the size of a grain of rice and is difficult to attack. Each file is encrypted
+            with AES-256-XTS, and the filesystem metadata is encrypted with AES-256-HEH.</p>
+
+            <p>File Based Encryption on GrapheneOS is always on, enforced, and cannot be switched
+            off or fall back to plaintext. The Titan M hardware security module has its own internal
+            timer which cannot be accessed from the host operating system, and can enforce
+            progressively lengthening delays between guesses at the password. On enough failed
+            guesses, the Titan M can increase the time between guesses to allow only one guess
+            per day. At this rate, to search through all the combinations of a four-digit pin code
```
inthewaves: pin code seems redundant
```suggestion
            per day. At this rate, to search through all the combinations of a four-digit PIN
```

static/faq.html
```diff
@@ -291,6 +294,55 @@ <h2 id="default-connections">
...
+            from 0000 to 9950 would take over 27 years, or for a relatively weak two-word passphrase
+            from a 2048-word list, eleven thousand years. GrapheneOS supports passphrases up to 64
+            characters long for extra security as a contingency plan in the event an exploit is
+            discovered for the Titan M that might allow the timer to be shortened.</p>
+
+            <h2 id="fbe-over-fde">
+                <a href="#fbe-over-fde">Why does GrapheneOS use FBE over FDE?</a>
+            </h2>
+            <p>Virtually all modern phones use file based encryption instead of full disk
```
inthewaves:
```suggestion
            <p>Virtually all modern phones use file-based encryption instead of full-disk
```

static/faq.html
```diff
@@ -291,6 +294,55 @@ <h2 id="default-connections">
...
+            encryption. Android 10 requires file based encryption. Utilizing file based encryption
```
inthewaves:
```suggestion
            encryption. Android 10 requires file-based encryption. Utilizing file-based encryption
```

static/faq.html
```diff
@@ -291,6 +294,55 @@ <h2 id="default-connections">
...
+            the token to the phone so it can in turn derive the keys to decrypt the files if it gets
+            the correct password.</p>
+
+            <p>This system ensures that an adversary in possession of your phone cannot simply
+            extract the SSD and connect it to a much more powerful system to run guessing attacks on
+            it offline, and instead must go through the specific HSM on the phone. The HSM is
+            about the size of a grain of rice and is difficult to attack. Each file is encrypted
+            with AES-256-XTS, and the filesystem metadata is encrypted with AES-256-HEH.</p>
+
+            <p>File Based Encryption on GrapheneOS is always on, enforced, and cannot be switched
```
inthewaves:
```suggestion
            <p>File-based encryption on GrapheneOS is always on, enforced, and cannot be switched
```

static/faq.html
```diff
@@ -291,6 +294,55 @@ <h2 id="default-connections">
...
+            encryption. Android 10 requires file based encryption. Utilizing file based encryption
+            has a number of advantages:
+                <ul>
+                    <li>It allows work profiles to be made more secure, since when the user is
+                    logged out, the keys can be removed from memory and remain encrypted.</li>
+                    <li>Apps that are written to use Strongbox Keymaster can eject their own keys
+                    from memory when the screen is locked, but still be able to function and append
+                    data to storage even while the data on the phone is locked and encrypted.</li>
+                    <li>The phone can handle updates even while locked and the data is at rest.</li>
+                    <li>File based encryption allows some features like direct boot, which many
```
inthewaves:
```suggestion
                    <li>File-based encryption allows some features like direct boot, which many
```

static/faq.html
```diff
@@ -291,6 +294,55 @@ <h2 id="default-connections">
...
+
+            <h2 id="fbe-over-fde">
+                <a href="#fbe-over-fde">Why does GrapheneOS use FBE over FDE?</a>
+            </h2>
+            <p>Virtually all modern phones use file based encryption instead of full disk
+            encryption. Android 10 requires file based encryption. Utilizing file based encryption
+            has a number of advantages:
+                <ul>
+                    <li>It allows work profiles to be made more secure, since when the user is
+                    logged out, the keys can be removed from memory and remain encrypted.</li>
```
inthewaves: I think this can be made more general as it applies also for user profiles, not just work profiles

static/faq.html
```diff
@@ -291,6 +294,55 @@ <h2 id="default-connections">
                 </li>
             </ul>
 
+            <h2 id="securing-data-at-rest">
+                <a href="#securing-data-at-rest">How does GrapheneOS secure data at rest?</a>
+            </h2>
+            <p>GrapheneOS makes full use of the hardware-bound encryption in modern Android devices.
+            The Titan M, a specially designed hardware security module on Pixel 3 handsets handles
+            validation of the user's pincode or password. Upon getting the correct password, the HSM
```
inthewaves:
```suggestion
            validation of the user's PIN or password. Upon getting the correct password, the HSM
```

static/faq.html
```diff
@@ -291,6 +294,55 @@ <h2 id="default-connections">
...
+            confines of the HSM, and from a hash of a two-kilobyte partition at the start of the
+            phone's SSD by means of a key derivation function. The HSM will only be able to release
+            the token to the phone so it can in turn derive the keys to decrypt the files if it gets
+            the correct password.</p>
+
+            <p>This system ensures that an adversary in possession of your phone cannot simply
+            extract the SSD and connect it to a much more powerful system to run guessing attacks on
+            it offline, and instead must go through the specific HSM on the phone. The HSM is
+            about the size of a grain of rice and is difficult to attack. Each file is encrypted
+            with AES-256-XTS, and the filesystem metadata is encrypted with AES-256-HEH.</p>
```
inthewaves, line 588: The vocab is a bit inconsistent; it jumps between "HSM" in this paragraph to "Titan M" in the next paragraph

static/faq.html
Peter-Easton: Sure. Let me go and fix that...

static/faq.html
Peter-Easton: Sure. I'll go and get that...

static/faq.html
Peter-Easton: Good thing you brought this up, since after you mentioned it I decided I'd give it a test using my own phone to see if it actually took more than 50 to make it from 30 seconds to a minute and up and up. Originally I was under the impression that it would behave by doubling the time interval, but I decided to experiment a little bit and test with a password, pincode, and pattern, and my results actually differed a bit from what I expected:

After the first five attempts the phone will display the number of incorrect password attempts in a dialogue box, disable the fingerprint sensor and reject all input for 30 seconds. After the next five attempts, the phone will display the number of incorrect password attempts in a dialogue box and reject all input for another 30 seconds. After each unsuccessful attempt at that point, it'll do the same.

I'll actually have to revisit this section, whoops.

static/faq.html
Peter-Easton: Should be fixed now!

static/faq.html
Peter-Easton: Okay! I dropped the word "work" profiles, how does it sound?

Peter-Easton:
Okay! I'm going to try to clean up the git history of this up a little bit, rebase and push this back. Wish me luck, hopefully it should save your edits.

static/faq.html
inthewaves: I think it should be "user profiles", since the OS uses "profile" in the strings (and also https://source.android.com/devices/tech/admin/multi-user). "Profiles" alone may be ambiguous

I think it would also be good to mention per-profile encryption keys as well, since it's not explicitly mentioned here. 

So, something like, "It allows user profiles to be made more secure, since each profile has their own encryption keys that get wiped from memory when logged out. Data in an inactive profile becomes at rest (i.e., the data becomes inaccessible, because the encryption key has to be reobtained from user authentication)." This would also clarify what we mean by "at rest."

There's more detail given recently on this topic: https://reddit.com/r/GrapheneOS/comments/f1b2l7/wipe_free_space/fh8lvk9?context=3

static/faq.html
thestinger: The Titan M is only involved for the initial unlock after rebooting. Make sure you're properly testing this.

static/faq.html
Peter-Easton: Sorry about that. I had some problems and this issue fell through the cracks. I'll fix that next push.

Peter-Easton: Should be fixed now.

static/faq.html
Peter-Easton: I followed the instructions and tested it to 61 results this time. The time delay starts to increment from 30 seconds to 60 seconds at 40 tries, after which the time lockout doubles to two minutes at 50 tries, then four minutes at 60 tries. I think that provided that it continues to work this way, then it would take 100 tries to reach one hour, and then from there after 150 tries it would reach one day between attempts. 

While I was at it, I also plugged a keyboard into my phone and attempted to enter the password, and it looks like GrapheneOS won't allow this which would make automated password spoofing even more difficult. 

Looks like it is working exactly as it is intended to, I just was mistaken and thought that it would be incrementing faster. I'll update the draft with this new information.

thestinger: It allows attaching a keyboard but only detects it at boot or when unlocked, by default.

Peter-Easton:
Hey guys. I'm going to close this one for a bit because it's been a long time and since then it's become a bit stale. I'll update it with a newer revision later, I've learned a bit since.

PR #50, opened 2020-02-18
fix typos, reword some phrases, and add extra info

inthewaves:
Also, mention IP68 water and dust resistance for Pixel 3/3 XL


static/faq.html
```diff
@@ -86,8 +86,8 @@ <h2 id="supported-devices">
             support repositories for the Android Open Source Project can simply be dropped into
             the source tree, with at most minor modifications within them to support GrapheneOS.
             In most cases, substantial work beyond that will be needed to bring the support up to
-            the same standards. For most devices, the hardware and firmware will prevent providing
-            a reasonably secure device, regardless of the work put into device support.</p>
+            the same standards. For most devices, the hardware and firmware will prevent them from
```
thestinger: I'd rather leave this part as it was, since this is changing the meaning in a way that wasn't intended.

inthewaves: done

PR #52, opened 2020-02-19
Add Q&A about cell tracking and silent sms

Peter-Easton:


static/faq.html
```diff
@@ -291,6 +293,22 @@ <h2 id="default-connections">
...
             </ul>
 
+            <h2 id="cellular-tracking">
+                <a href="#cellular-tracking">What does GrapheneOS do about cellular tracking and
+                silent SMS?</a>
+            </h2>
+
+            <p>GrapheneOS always considers the network to be hostile. Merely getting a silent SMS is
+            not a good indicator of being targeted by your cell carrier, police or government
+            because <i>anyone on the cell network can send them</i>. Cellular triangulation will
```
thestinger: You should use `<em></em>`, not `<i></i>`. Similarly, use `<strong></strong>` rather than `<b></b>`.

thestinger (changes requested):

static/faq.html
```diff
@@ -291,6 +293,22 @@ <h2 id="default-connections">
...
+            happen regardless of whether or not SMS texts are being sent or recieved by the phone,
+            and they don't have to even be a silent SMS; it wouldn't be much different than
+            recieving an unsolicited spam SMS!</p>
+
+            <p>Activating airplane mode will switch the cell radio off entirely. While airplane mode
```
thestinger: Should clarify and state that it fully disables cellular radio transmit / receive capabilities. The baseband processor provides more than the cellular radio. On the currently supported devices, it provides both the Wi-Fi and cellular radios with separately sandboxed code for each.

thestinger (changes requested):

static/faq.html
Peter-Easton: Whoops! Okay! Let me go and fix that...

static/faq.html
Peter-Easton: Woah, whew, good to know, I'll go and fix that...

static/faq.html
thestinger: https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-em-element vs. https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-i-element

https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-strong-element vs. https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-b-element

Essentially, em and strong are for the semantic meanings associated with those, and i and b are for more subtle cases without those clear semantic meanings but rather a more subtle reason. For a long time, i and b were deprecated, but they came up with new meanings for them giving them purpose again and they aren't deprecated anymore.

thestinger: It matters for accessibility, since accessibility tools can read the text in a different way for em and strong, but the others are more subtle.

Peter-Easton: I think this may appear in a couple other places, I'll go get them too.

PR #55, opened 2020-02-19
Add Q&A about Signal Private Messenger

Peter-Easton:
This one adds a "Day To Day" category.


thestinger:
Documenting app compatibility and recommended is currently largely outside the scope of the guides. This should be done in a dedicated place. Web browsing is covered because we provide a web browser.

PR #57, opened 2020-02-20
Add Q&A about what 'not for general end users' means

Peter-Easton:


thestinger:
It's no longer being described as not production ready. It's just not yet aimed at usage by most end users until the app repository need is addressed, among other things.

thestinger (changes requested):
Can you reword this based on that?

Peter-Easton:
Sure. Will change it!

Peter-Easton:
There we go. I left most of it the same but I changed it to about being aimed at general end users.

thestinger (dismissed):

thestinger (changes requested):
I think you need to make it much clearer that GrapheneOS is intended for this, but isn't ready yet. This makes it sound this it's not intended for this use case, rather than this being under development.

Peter-Easton:
I think I fixed it, how's it sound now?

static/faq.html
```diff
@@ -668,6 +670,30 @@ <h3 id="updates-sideloading">
...
+            security-focused application store of specially curated first-party apps chosen
+            specifically for technical merit, with dedicated maintainers.</p>
+
+            <p>Since a dedicated, secure application store of specially curated first-party apps has
+            not yet been finished, this places the onus on the user to source and update their own
+            apps as an interim measure. Downloading and installing apps from sites run by other
+            parties on the Internet places the burden to find genuine apps on the user's judgement,
+            which is antithetical to good security design. Relying on trusted third parties that
+            use a centralized application signing model such as F-Droid runs contrary to the Android
+            security model. This choice to describe GrapheneOS as not being for most end users is
```
thestinger: Can you say not yet here?

thestinger (changes requested):

static/faq.html
Peter-Easton: Okay, one moment...

thestinger:
I don't want to include this.

PR #59, opened 2020-02-20
Add Q&A about vpn support

Peter-Easton:


thestinger:
This should probably mention the built-in VPN support in the OS, the always-on VPN option and the option to block connections not going via the configured VPN to prevent leaks if the VPN connection drops, etc.

Peter-Easton:
Got it. Let me go and add it!

static/faq.html
```diff
@@ -361,6 +367,23 @@ <h3 id="cellular-tracking">
...
+
+            <h3 id="vpn-support">
+                <a href="#vpn-support">Does GrapheneOS support VPNs and VPN apps?</a>
+            </h3>
+
+            <p>Yes. Orbot, OpenVPN for Android, the userspace implementation of WireGuard, and the
+            Private Internet Access VPN app have all been tested and reported working. VPNs can be
+            configured under Settings -> Network & Internet -> Advanced -> VPN.</p>
+
+            <p>GrapheneOS has built-in VPN support included in the operating system and has an
```
thestinger: Built-in VPN support should be the first paragraph, followed by app support, and then always-on VPN support. Always-on VPN support should be a separate paragraph. As is, this is going to be confusing, since it makes it seem like always-on VPN access is only for built-in VPN support.

thestinger (changes requested):

static/faq.html
Peter-Easton: Okay! I've tried reordering it for this one, looks like it made it through after a rebase.

PR #60, opened 2020-02-21
Add Q&A on the difficulty of the installation

Peter-Easton:


thestinger:
I don't want to include this, particularly the comparison to LineageOS, but in general I don't think it is desirable.

People who want to make the installation process easier can work on WebUSB-based installation.

PR #62, opened 2020-02-21
Add link to Reddit thread "OS security, iOS vs GrapheneOS vs Stock An…

Peter-Easton:
…droid"


renlord:
the answers from the reddit link should be recorded here instead.

thestinger:
Yeah, I'm going to close this for the time being. Information like this needs to be actively maintained so we can't just link to comments on Reddit.

PR #66, opened 2020-02-24
Add "Back To Top" links on all currently existant answers.

Peter-Easton:
I included the up arrow for this but I can remove it if it doesn't render correctly.


thestinger:
I'd rather not include this kind of navigation, at least right now.

PR #69, opened 2020-02-27
Add Q&A about MicroG and signature spoofing

Peter-Easton:


Peter-Easton:
This one's also been explained adequately on the website since under the heading about not using Google services, so it's technically redundant at this point, I'm going to close it from my end, but I'll keep it around just in case we need it later.

PR #70, opened 2020-02-27
Update donate page with new Github sponsorship information.

Peter-Easton:


thestinger:
Can you make this consistent? "5 to 100"

Peter-Easton:
Got it, now they're all in long form since the "five thousand" part was also in long form. All three of them should match now.

PR #74, opened 2020-03-09
Updated grapheneos.org design using material lite

wanwari:
Updated grapheneos.org design using material lite. Minimal changes done to actual structure and content of the website. Example of new design can be seen below

[Example 1](https://i.imgur.com/3y7S135.png)
[Example 2](https://i.imgur.com/58e1HUg.png)
[Example 3](https://i.imgur.com/VApiArO.png)

### DIFF
--- a/static/build.html
+++ b/static/build.html
@@ -29,9 +29,9 @@
     <body>
         <nav>
             <ul>
-                <li><a href="/">GrapheneOS</a></li>
+                <li class="first"><a href="/">GrapheneOS</a></li>
                 <li><a href="./install.html">Install</a></li>
-                <li class="active"><a href="./build.html">Build</a></li>
+                <li><a href="./build.html" class="first">Build</a></li>
                 <li><a href="./usage.html">Usage</a></li>
                 <li><a href="./faq.html">FAQ</a></li>
                 <li><a href="./releases.html">Releases</a></li>
--- a/static/build.html
+++ b/static/build.html
@@ -2,40 +2,42 @@
 <html lang="en" prefix="og: http://ogp.me/ns#">
     <head>
         <meta charset="utf-8"/>
-        <title>Build | GrapheneOS</title>
-        <meta name="description" content="Building instructions for GrapheneOS, a security and privacy focused mobile OS with Android app compatibility."/>
+        <title>GrapheneOS</title>
+        <meta name="description" content="GrapheneOS is a security and privacy focused mobile OS with Android app compatibility."/>
         <meta name="theme-color" content="#212121"/>
         <meta name="msapplication-TileColor" content="#ffffff"/>
         <meta name="viewport" content="width=device-width, initial-scale=1"/>
         <meta name="twitter:site" content="@GrapheneOS"/>
         <meta name="twitter:creator" content="@GrapheneOS"/>
-        <meta property="og:title" content="GrapheneOS build documentation"/>
-        <meta property="og:description" content="Building instructions for GrapheneOS, a security and privacy focused mobile OS with Android app compatibility."/>
+        <meta property="og:title" content="GrapheneOS"/>
+        <meta property="og:description" content="GrapheneOS is a security and privacy focused mobile OS with Android app compatibility."/>
         <meta property="og:type" content="website"/>
         <meta property="og:image" content="https://grapheneos.org/opengraph.png"/>
         <meta property="og:image:width" content="512"/>
         <meta property="og:image:height" content="512"/>
         <meta property="og:image:alt" content="GrapheneOS logo"/>
-        <meta property="og:url" content="https://grapheneos.org/build"/>
+        <meta property="og:url" content="https://grapheneos.org/"/>
         <meta property="og:site_name" content="GrapheneOS"/>
         <link rel="icon" type="image/vnd.microsoft.icon" href="/favicon.ico"/>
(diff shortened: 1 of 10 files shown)


static/build.html
```diff
@@ -2,32 +2,34 @@
...
         <meta property="og:image" content="https://grapheneos.org/opengraph.png"/>
         <meta property="og:image:width" content="512"/>
         <meta property="og:image:height" content="512"/>
         <meta property="og:image:alt" content="GrapheneOS logo"/>
-        <meta property="og:url" content="https://grapheneos.org/build"/>
+        <meta property="og:url" content="https://grapheneos.org/"/>
         <meta property="og:site_name" content="GrapheneOS"/>
         <link rel="icon" type="image/vnd.microsoft.icon" href="/favicon.ico"/>
         <link rel="mask-icon" href="/mask-icon.svg" color="#1a1a1a"/>
+        <link rel="stylesheet" href="https://fonts.googleapis.com/icon?family=Material+Icons">
```
thestinger, line 23: I don't want to include any external resources. Everything needs to be served as first-party content.

static/build.html
```diff
@@ -2,32 +2,34 @@
 <html lang="en" prefix="og: http://ogp.me/ns#">
     <head>
         <meta charset="utf-8"/>
-        <title>Build | GrapheneOS</title>
-        <meta name="description" content="Building instructions for GrapheneOS, a security and privacy focused mobile OS with Android app compatibility."/>
+        <title>GrapheneOS</title>
```
thestinger, line 5: It seems that you're unintentionally clobbering each of the metadata sections with the one for the index page.

static/build.html
```diff
@@ -306,7 +308,7 @@ <h2 id="kernel">
             <pre>cd kernel/google/crosshatch
 git submodule sync
 git submodule update --init
-./build.sh blueline</pre>
+/build.sh blueline</pre>
```
thestinger, line 311: These changes to paths look like a mistake.

static/donate.html
```diff
@@ -58,7 +60,7 @@ <h2 id="bitcoin">
             <h2 id="paypal">
                 <a href="#paypal">PayPal</a>
             </h2>
-            <p>PayPal donations can be sent to danielmicay@gmail.com. Please mention GrapheneOS in
+            <p>PayPal donations can be sent to <code>danielmicay@gmail.com</code>. Please mention GrapheneOS in
```
thestinger, line 63: I don't think email addresses should be styled as code.

static/contact.html
```diff
@@ -56,8 +58,8 @@ <h2 id="community">
             </h2>
             <p>The main forum currently used to discuss the project is the official
             <a href="https://reddit.com/r/GrapheneOS">/r/GrapheneOS subreddit</a> on Reddit.</p>
-            <p>The official IRC channel is #grapheneos on irc.freenode.net. The channel is bridged
-            to Matrix and is available at #grapheneos:matrix.org for Matrix users.</p>
+            <p>The official IRC channel is <code>#grapheneos</code> on <code>irc.freenode.net</code>. The channel is bridged
```
thestinger, line 61: I'm not sure that I want to change how this is styled. It probably belongs in a separate pull request either way.

static/build.html
```diff
@@ -2,32 +2,34 @@
...
         <meta property="og:image:width" content="512"/>
         <meta property="og:image:height" content="512"/>
         <meta property="og:image:alt" content="GrapheneOS logo"/>
-        <meta property="og:url" content="https://grapheneos.org/build"/>
+        <meta property="og:url" content="https://grapheneos.org/"/>
         <meta property="og:site_name" content="GrapheneOS"/>
         <link rel="icon" type="image/vnd.microsoft.icon" href="/favicon.ico"/>
         <link rel="mask-icon" href="/mask-icon.svg" color="#1a1a1a"/>
+        <link rel="stylesheet" href="https://fonts.googleapis.com/icon?family=Material+Icons">
+        <link rel="stylesheet" href="https://code.getmdl.io/1.3.0/material.indigo-pink.min.css">
```
thestinger, line 24: I want to keep the current theme matching the theme used by GrapheneOS developed apps, which is the default Material theme. A pink theme doesn't fit GrapheneOS.

static/grapheneos.css
```diff
@@ -52,31 +52,42 @@ nav ul {
 }
 
 nav {
-    background-color: #212121; /* grey900 */
-    box-shadow: 0 2px 2px 0 rgba(0, 0, 0, 0.14), 0 1px 5px 0 rgba(0, 0, 0, 0.12), 0 3px 1px -2px rgba(0, 0, 0, 0.2);
+    background-color: #c2185b;
+    -webkit-box-shadow: 0px 1px 5px 0px rgba(0,0,0,0.75);
```
thestinger, line 56: Should avoid prefixes when they aren't required for current generation browsers.

thestinger (changes requested):
I don't necessarily want to incorporate something like this, but the review comments would need to be addressed for this implementation of it to be considered. I'd need to be given compelling reasons to incorporate any kind of framework along with having a clear understanding of the advantages / drawbacks and how well the code is maintained, etc.

static/build.html
wanwari: My apologies, I assumed the material lite framework was a framework that wanted to be used based on the comments in the grapheneos.css file, however I can now understand that it is just the theme and not the actual framework.

PR #76, opened 2020-03-26
Clarify Q&A about what legacy means under older devices

Peter-Easton:


static/faq.html
```diff
@@ -254,6 +254,11 @@ <h3 id="legacy-devices">
             project ends up wanting to drop devices for this reason but has always kept them going
             until the end-of-life date to provide more time for people to migrate.</p>
 
+            <p>If a device is marked "legacy", this indicates that it no longer meets the minimum
+            standard for support now, even though it met them at an earlier point in time. Legacy
+            devices are in the process of being depreciated, and support for them will likely
+            cease soon.</p>
```
thestinger, line 260: It's not about when support will end.

thestinger (changes requested):

thestinger:
There aren't specific rules about what this means, and I don't think it really needs to be explained. It's a way of deterring people from buying past generation devices for various reasons.

PR #82, opened 2020-04-10
Clarify USB debugging is not required to install

shortcak:
There are people recommending that USB debugging be enabled for installing, but just using fastboot without it enabled is sufficient. Some guides (even though you shouldn't follow external guides) say this so that users can use an adb command to reboot the device into bootloader (some also recommend USB debugging it because it "ensures a proper connection between your devices")

### DIFF
--- a/static/install.html
+++ b/static/install.html
@@ -163,7 +163,11 @@ Installed as /home/username/downloads/platform-tools/fastboot</pre>
                 <a href="#unlocking-the-bootloader">Unlocking the bootloader</a>
             </h2>
             <p>First, boot into the bootloader interface. You can do this by turning off the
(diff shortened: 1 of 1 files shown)


thestinger:
It doesn't matter if people do this since it's about to be replaced by a whole new OS though. Flashing the OS from the host trusts that host too, which is covered in the guide already.

thestinger:
I don't really want to make the guide longer / more complicated to cover things people shouldn't do. I'd be fine with a simpler, smaller note that ADB is not required for the installation process but as is this change doesn't make sense since the existing OS is replaced, the data is all wiped, and the host is trusted regardless (other than attestation, but an unpaired attestation is weak).

PR #84, opened 2020-04-16
Add mention of IRC logging to contact.

Peter-Easton:


Peter-Easton:
I'd like to clean up the topic on the IRC channel and rather than have eight links in the topic channel, have only one. Should we try linking it to this page instead? I can reorganize it a bit.

PR #87, opened 2020-04-26
add more information to install page

theaeonsolution:
Added information to help users that get stuck on the OEM unlocking part because they got a carrier locked device and wanted to share my insight if they have a device that can't be unlocked. I also wanted to add that Arch Linux is great but many people may not know how to set that up. Adding Manjaro as a friendly alternative might help speed up the process of getting GrapheneOS on their device(s).

### DIFF
--- a/static/install.html
+++ b/static/install.html
@@ -79,13 +79,12 @@
             <p>You need one of the officially supported devices. To make sure that the device can
             be unlocked to install GrapheneOS, avoid carrier variants of the devices. Carrier
             variants of Pixels use the same stock OS and firmware with a non-zero carrier id
-            flashed onto the persist partition in the factory. The carrier id activates
-            carrier-specific configuration in the stock OS including disabling carrier and
+            flashed onto the persist partition in the factory state. The carrier id activates
+            the carrier-specific configuration in the stock OS including disabling carrier and
             bootloader unlocking. The carrier may be able to remotely disable this, but their
             support staff may not be aware and they probably won't do it. Get a carrier agnostic
             device to avoid the risk and potential hassle. If you CAN figure out a way to unlock a
-            carrier device, it isn't a problem as GrapheneOS can just ignore the carrier id and
-            it's otherwise the same.</p>
+            carrier device, installing GrapheneOS won't be a problem.</p>
             <p>It's best practice to update the stock OS on the device to make sure it's running
             the latest firmware before proceeding with these instructions. This avoids running
             into bugs, missing features or other differences in older firmware versions. Early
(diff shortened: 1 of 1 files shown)


static/install.html
```diff
@@ -106,6 +107,10 @@ <h3 id="obtaining-fastboot">
                     tools, although they aren't necessary to install) and
                     <code>android-udev</code> for udev rules to make fastboot / adb work in local
                     sessions without root</li>
+                <li>Manjaro Linux: is a friendly and open-source Linux distribution based
```
thestinger: Can you avoid describing the distribution here? I don't want to advertise it.

static/install.html
```diff
@@ -158,7 +163,9 @@ <h3 id="obtaining-signify">
             <h2 id="enabling-oem-unlocking">
                 <a href="#enabling-oem-unlocking">Enabling OEM unlocking</a>
             </h2>
-            <p>OEM unlocking needs to be enabled from within the operating system.</p>
+            <p>OEM unlocking needs to be enabled from within the operating system.
```
thestinger: It's usually grayed out because of Factory Reset Protection in the stock OS which requires being able to connect to the internet to change this.

static/install.html
```diff
@@ -191,7 +198,7 @@ <h2 id="obtaining-factory-images">
             <a href="https://www.reddit.com/r/GrapheneOS/comments/c7gb3f/grapheneos_factory_images_are_now_signed_with/esewpm9">the /u/GrapheneOS
             Reddit account</a> and <a href="https://github.com/GrapheneOS/releases.grapheneos.org/blob/master/static/factory.pub">is available on GitHub</a>.
             When the current signing key is replaced, the new key will be signed with it.</p>
-            <p>Download the factory images for the device from <a href="/releases">the releases
+            <p>Download both factory images(*.zip and *.zip.sig) for the device from <a href="/releases">the releases
```
thestinger: I can see what you're trying to do here but the implementation of the change doesn't make sense. The signature isn't another set of factory images. I'll change this and you can just omit the change.

thestinger (changes requested):
Forgot to mark changes as being requested. GitHub's review interface is a pain.

thestinger:
I made some changes myself so you'll need to rebase this.

static/install.html
theaeonsolution: Note taken, thank you for the feedback.

theaeonsolution:
Added and omitted the necessary information after the rebase.

static/install.html
theaeonsolution: I missed this change in my update. Please ignore the patch.

theaeonsolution:
Stale pull request

PR #88, opened 2020-04-27
Add back old writeup on network and sensors permission to usage guide

Peter-Easton:
I copied some of the old information from the old legacy documents.


Peter-Easton:
Hey guys. I'm going to close this one because it's been a long time and since then it's become a bit stale. I'll update it with a newer revision later, I've learned a bit since.

PR #89, opened 2020-04-27
Move and clarify information about hardware IDs to usage guide

Peter-Easton:
I copied some of the old information from the old legacy documents. I also decided for the sake of brevity to start moving some stuff out of the FAQ since I've been using the FAQ as a bit of a dumping ground for more technical information that should be in its proper place.

### DIFF
--- a/static/usage.html
+++ b/static/usage.html
@@ -63,6 +63,13 @@
                     </ul>
                 </li>
                 <li><a href="#web-browsing">Web browsing</a></li>
+                <li>
+                    <a href="#background-permissions">Background permissions</a>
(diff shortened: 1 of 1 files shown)


thestinger (changes requested):
I'd rather leave this in the FAQ since there's nothing for users to actually do with it. I think the difference should be that the usage guide is a guide on how to use the OS and the FAQ provides background information and addresses common questions / concerns. Sometimes it may make sense to split information between them.

Peter-Easton:
Sure. Let me see if I can figure out something that reads more elegantly...

PR #90, opened 2020-04-27
Add messaging recommendations from old documents

Peter-Easton:
Update messaging recommendations for year 2020

Thought I'd do this so we could start moving some of it out of the FAQ and it was already in the old usage guide.


gosh-darn:
I think the use of WhatsApp should be discouraged. It is closed source and currently in the hands of Facebook, if I'm not wrong.

thestinger:
Don't really want to have these kinds of app recommendations as part of the guides anymore. It could be done elsewhere, but it'll need to be done from scratch.

PR #96, opened 2020-05-04
fix Windows command type and clarify expected result / steps after

pjvleeuwen:
FIX COMMAND : otherwise PS will report "flash-all : The term 'flash-all' is not recognized as the name of a cmdlet, function, script file, or..."
CLARIFY STEPS AFTER : apparently I was able to fail to understand the meaning of the instructions until some iterations, so attempting to make it easier to understand for new users like me

### DIFF
--- a/static/install.html
+++ b/static/install.html
@@ -279,17 +279,14 @@ cd crosshatch-qq2a.200405.005</pre>
 
             <p>Wait for the flashing process to complete (message about not formatting 'userdata' 
             and 'metadata' can be expected).</p>
-            
+
             <p>On current generation devices like the Pixel 3, Pixel 3 XL, Pixel 3a and Pixel 3a
             XL the script ends with the fastbootd menu on the phone. On other devices it might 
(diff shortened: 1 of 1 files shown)


thestinger:
Using `flash-all` was intended since `$env:pathext` includes `.bat` by default which makes this work. It makes sense to specify it explicit in case `.bat` isn't included in `$env:pathext`.

thestinger:
I haven't quite addressed all this but I'm going to close this since it'll heavily conflict now and I don't want all the proposed changes. I appreciate the testing and reports.

pjvleeuwen:
> Using `flash-all` was intended since `$env:pathext` includes `.bat` by default which makes this work. It makes sense to specify it explicit in case `.bat` isn't included in `$env:pathext`.

In my case it did not work as it was (`flash-all`). I did not change the `pathext`, but maybe PS expected the `./` like bash (I never use PS anyway). Either way, as it is currently in the guide it works for sure, thanks :-)

PR #100, opened 2020-05-09
move Vanadium configuration information on build page

theaeonsolution:


static/build.html
```diff
@@ -656,21 +656,18 @@ <h3 id="browser-and-webview">
...
 
-            <p>You can obtain the proper configuration from the
-            <a href="https://github.com/GrapheneOS/Vanadium">
-            GrapheneOS Vanadium repository</a> in <code>args.gn</code> including the correct
-            version.</p>
-
             <pre>git clone https://github.com/GrapheneOS/Vanadium.git
-cd Vanadium
-git checkout $CORRECT_BRANCH_OR_TAG</pre>
+            cd Vanadium
```
thestinger: This whitespace shouldn't be added since the pre tag will include it.

thestinger (changes requested):

static/build.html
theaeonsolution: Update made. I really thought I could squash commits on the site, I will try to make less noise with changes and work locally before pushing commits.

static/build.html
```diff
@@ -669,8 +664,10 @@ <h3 id="browser-and-webview">
 
             <pre>fetch --nohooks android</pre>
 
-            <p>Sync to the latest stable release for Android (replace $VERSION with the correct
-            value):</p>
+            <p>You can obtain the proper configuration from the <a href="https://github.com/GrapheneOS/Vanadium">
```
thestinger, line 667: You should probably move this down a bit further if you want to move it like this. I don't think it makes more sense here. It should go with the configuration step.

thestinger (changes requested):

static/build.html
theaeonsolution: Thanks, let me review.

PR #101, opened 2020-05-17
update build page wording

theaeonsolution:


thestinger:
I prefer the existing wording in at least several of these cases, so I can't merge this. You would be better off submitting these kinds of subjective changes in the smallest possible commits because if if I don't like even just one of the changes I can't merge it.

thestinger:
Some of this is changing the meaning of what I wrote in ways that aren't intended, which would make some of these things less accurate and specific.

PR #117, opened 2020-07-11
Add information on installing debug builds

Peter-Easton:
This is so far just the quickest and most convenient way to get a debug build onto the phone. So far it's worked for me on the Pixel 4.

### DIFF
--- a/static/build.html
+++ b/static/build.html
@@ -69,7 +69,12 @@
                         <li><a href="#reproducible-builds">Reproducible builds</a></li>
                         <li><a href="#extracting-vendor-files-for-pixel-devices">Extracting vendor files for Pixel devices</a></li>
                         <li><a href="#building">Building</a></li>
-                        <li><a href="#faster-builds-for-development-use-only">Faster builds for development use only</a></li>
+                        <li>
+                            <a href="#faster-builds-for-development-use-only">Faster builds for development use only</a>
+                            <ul>
+                                <li><a href="#installing-debug-builds">Installing debug builds</a></li>
+                            </ul>
+                        </li>
                         <li>
                             <a href="#generating-release-signing-keys">Generating release signing keys</a>
                             <ul>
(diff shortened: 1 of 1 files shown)


static/build.html
```diff
@@ -534,6 +539,25 @@ <h3 id="faster-builds-for-development-use-only">
             lot more sense to test it with proper signing keys rather than the default public test
             keys.</p>
 
+
+            <h4 id="installing-debug-builds">
+                <a href="#installing-debug-builds">Installing debug builds</a>
+            </h4>
+
+            <p>The release script will not work for debug or userdebug builds, which are intended
```
thestinger, line 547: This isn't accurate - you can do signed userdebug builds.

thestinger, line 547: This isn't accurate - you can do signed userdebug builds.

static/build.html
```diff
@@ -534,6 +539,25 @@ <h3 id="faster-builds-for-development-use-only">
...
+            for testing and debugging only. To install a debug build on a device, after finishing
+            the debug or userdebug build and unlocking the bootloader, it is necessary to use
+            fastboot and set the environment variable with the following command sequence (replacing
+            <code>crosshatch</code> with the appropriate device codename):</p>
+
+            <pre>cd out/target/product/crosshatch
+ANDROID_PRODUCT_OUT=`pwd` fastboot flashall -w</pre>
```
thestinger, line 554: This won't flash everything and may lead to people having problems.

static/build.html
```diff
@@ -534,6 +539,25 @@ <h3 id="faster-builds-for-development-use-only">
...
+
+            <p>Wait for the command to run to completion. Note that verified boot is <em>not</em>
```
thestinger, line 556: This isn't accurate.

thestinger (changes requested):

thestinger:
I can't merge this because it's not correct.

Peter-Easton:
Ack, whoops, sorry about that. I'll go and try to fix some things and try again later.

thestinger:
@Peter-Easton flashall won't flash the firmware images which is required if they need to be updated. Also, nothing prevents making a `userdebug` variant build and using target-files-package + signing it with the release script. The difference between `user` and `userdebug` is about whether debugging features like `adb root`, `su` for `adb shell`, all apps being debuggable, etc. are available

thestinger:
It usually makes sense to user `userdebug` for development builds. It usually makes sense not to sign development builds with release keys. It doesn't make much sense to generate factory images / over-the-air updates for builds that aren't signed with release keys. It doesn't mean that those things are inherently tied together. You *could* use the test key signed target-files-package for a userdebug build to generate factory images and a test key signed over-the-air update. That's not really ever going to be useful.

For running the Compatibility Test Suite, to pass all the tests you need a `user` build signed with release keys and the bootloader needs to be locked, otherwise some tests are going to fail.

thestinger:
And for example, if you only make a kernel code change, you can get away with only flashing `boot` as long as verified boot is disabled (bootloader unlocked). It will trigger a warning in the OS due to a fingerprint mismatch if the fingerprint has changed but that's expected and unimportant for testing - and the fingerprint won't change for a regular incremental build. The same goes for flashing only system, only vendor, etc. If you know what you're doing you can save a lot of time by only flashing whatever changed.

On a userdebug build, it's possible to use `adb root` and make use of faster ways of testing changes. A userdebug build supports doing `adb disable-verity` and then using `adb sync` to sync file changes to high level OS images like product, system, vendor, etc. They won't pass verification anymore and it doesn't sync low-level images like vbmeta/boot/dtbo which are not OS filesystems.

These things matter a lot for quick development cycles. It has to be properly and accurately explained or it does more harm than good to talk about it in the documentation.

thestinger:
There is still verified boot support without signing the build with release keys. Everything is still signed before signing a build with release keys. It's signed with the AOSP test keys which have public private keys. If you really wanted, you could create and flash an avb_pkmd.bin for the AOSP AVB test key that's being used and lock the bootloader to test verified boot without needing to sign with release keys. Nothing prevents using factory images / over-the-air updates with a test key signed build. This just doesn't usually make any sense.

PR #122, opened 2020-08-22
wording

theaeonsolution:


thestinger:
The current wording is current - it either has to say companies and other organizations or just organizations. Companies are organizations.

PR #140, opened 2020-11-19
Add FAQ about browser fingerprinting

Peter-Easton:


Peter-Easton:
I'm going to close this one because the site has changed significantly since, but I'll reopen it later as a different PR.

PR #146, opened 2020-12-21
Added scrollbar to site menu

floromancy:


static/grapheneos.css
```diff
@@ -139,6 +139,24 @@ footer a, footer a:visited {
     padding: 0;
     list-style-type: none;
 }
+/* Scrollbar  */
+::-webkit-scrollbar {
+    -webkit-appearance: none;
+ }
+ 
+ ::-webkit-scrollbar:horizontal {
```
thestinger, line 147: Can you remove these single spaces at the start of these lines? I think that's what the validation is unhappy about.

PR #148, opened 2020-12-22
Added visible scrollbar for menu on mobile devices

loimography:


thestinger:
This works for always showing the scrollbar, but I don't really like how it looks.

thestinger:
I didn't like how this looked and we'll need to figure out a nice way of doing it. I would prefer simply always making the standard scrollbar visible but sadly that doesn't appear to be possible.

PR #155, opened 2021-01-12
FAQ: Add clarification that profiles use different keys even if passw…

Peter-Easton:
…ords are reused.


thestinger:
Applied with my own wording changes. Thanks.

PR #168, opened 2021-01-30
Add periods

madaidan:
It's grammatically incorrect otherwise.


h4xor666:
This actually isn't true. Because it is a list, as well as not complete sentences throughout the list, it does not need periods.

madaidan:
That's not true.

https://www.probizwriters.com/PBW-blog/index.php/punctuation-of-bullet-lists/

thestinger:
This might be desirable but heavily conflicts at this point.

PR #176, opened 2021-02-09
use correct codenames for pixel 4a 5G and pixel 5

shagu:
Use correct codenames for Pixel 4a 5G and Pixel 5. In the current version of the document, both are swapped.
The Pixel 5 should be `redfin`, Pixel 4a 5G should be `bramble`.

References: 
- [Wikipedia Pixel 4a 5G](https://en.wikipedia.org/wiki/Pixel_4a)
- [Wikipedia Pixel 5](https://en.wikipedia.org/wiki/Pixel_5)
- [source.android.com](https://source.android.com/setup/build/building-kernels) *(search for bramble or redfin there)*

### DIFF
--- a/static/build.html
+++ b/static/build.html
@@ -142,8 +142,8 @@
                         <li>aosp_coral (Pixel 4 XL)</li>
                         <li>aosp_flame (Pixel 4)</li>
                         <li>aosp_sunfish (Pixel 4a)</li>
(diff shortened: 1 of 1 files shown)


thestinger:
Don't need to provide references for it. It was only flipped in the build guide by accident. It's done correctly everywhere else.

PR #177, opened 2021-02-12
Clarify when/how to update driver

ghost:


thestinger:
@marcojanotta The new table of contents entry also needs to be added for the 2nd page.

PR #186, opened 2021-02-21
Pixel 4a 5G & Pixel 5 don't have Pixel Neural Core

kopolee11:
A change to indicate that now only Pixel 3 and Pixel 4 have Pixel Visual Core / Pixel Neural Core.

### DIFF
--- a/static/usage.html
+++ b/static/usage.html
@@ -313,9 +313,10 @@
                 it's going to be replaced. In the short term, there are other apps available providing
                 more capabilities and better support for taking advantage of the hardware.</p>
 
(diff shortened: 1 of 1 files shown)


thestinger:
You're misinterpreting the meaning of the slash. It's included there to mean *either* not *and*. This makes it a lot more confusing and I don't think it's a good way of updating it to include the Pixel 4a (5G) and Pixel 5. The reason the Pixel Neural Core is gone is because of improved Qualcomm SoC functionality.

thestinger:
This makes it a lot more confusing and I don't think it's a good way of updating it to include the Pixel 4a (5G) and Pixel 5. The reason the Pixel Neural Core is gone is because of improved Qualcomm SoC functionality. The section probably needs to be heavily rewritten.

PR #193, opened 2021-03-15
Clarify enabling developer options.

Forbo:


thestinger:
This needs to be kept in sync across both guides.

Forbo:
I've done the same for the CLI guide now as well.

thestinger:
The tense here doesn't matter the rest of the sense. It also doesn't really match what's said afterwards.

thestinger:
Used this wording instead: https://github.com/GrapheneOS/grapheneos.org/commit/2a08a0cfea73cd2c5e2086e80847110261b5ccd4.

PR #197, opened 2021-03-15
Continue job if apt updates fail

flawedworld:


thestinger:
Don't want to take this approach.

PR #199, opened 2021-03-28
Add dark color scheme

ghost:


thestinger:
@marcojanotta Can you change this for AttestationServer too? I want to keep attestation.app and grapheneos.org in sync.

PR #201, opened 2021-04-01
Create an alert if user tries to leave during download or installation.

flawedworld:
Solves https://github.com/GrapheneOS/os_issue_tracker/issues/469

### DIFF
--- a/static/js/web-install.js
+++ b/static/js/web-install.js
@@ -7,6 +7,8 @@ const RELEASES_URL = "https://releases.grapheneos.org";
 const CACHE_DB_NAME = "BlobStore";
 const CACHE_DB_VERSION = 1;
 
+let safeToLeave = true;
+
 // This wraps XHR because getting progress updates with fetch() is overly complicated.
 function fetchBlobWithProgress(url, onProgress) {
     let xhr = new XMLHttpRequest();
@@ -155,11 +157,16 @@ async function downloadRelease(setProgress) {
     let latestZip = await getLatestRelease();
 
     // Download and cache the zip as a blob
+    safeToLeave = false;
     setProgress(`Downloading ${latestZip}...`);
     await blobStore.init();
-    await blobStore.download(`${RELEASES_URL}/${latestZip}`, (progress) => {
-        setProgress(`Downloading ${latestZip}...`, progress);
-    });
+    try {
+        await blobStore.download(`${RELEASES_URL}/${latestZip}`, (progress) => {
+            setProgress(`Downloading ${latestZip}...`, progress);
+        });
+    } finally {
+        safeToLeave = true;
+    }
     setProgress(`Downloaded ${latestZip} release.`, 1.0);
 }
 
@@ -196,13 +203,18 @@ async function flashRelease(setProgress) {
     }
 
     setProgress("Flashing release...");
-    await device.flashFactoryZip(blob, true, reconnectCallback,
-        (action, item, progress) => {
-            let userAction = fastboot.USER_ACTION_MAP[action];
-            let userItem = item === "avb_custom_key" ? "verified boot key" : item;
-            setProgress(`${userAction} ${userItem}...`, progress);
-        }
-    );
+    safeToLeave = false;
+    try {
+        await device.flashFactoryZip(blob, true, reconnectCallback,
+            (action, item, progress) => {
+                let userAction = fastboot.USER_ACTION_MAP[action];
+                let userItem = item === "avb_custom_key" ? "verified boot key" : item;
+                setProgress(`${userAction} ${userItem}...`, progress);
+            }
+        );
+    } finally {
+        safeToLeave = true;
+    }
 
     return `Flashed ${latestZip} to device.`;
 }
@@ -298,5 +310,13 @@ if ("usb" in navigator) {
 } else {
     console.log("WebUSB unavailable");
 }
+
+// This will create an alert box to stop the user from leaving the page during actions
+window.addEventListener("beforeunload", event => {
+    if (!safeToLeave) {
+        console.log("User tried to leave the page whilst unsafe to leave!");
+        event.returnValue = "";
+    }
+});
 
 // @license-end


static/install/web.html
```diff
@@ -28,7 +28,7 @@
         <link rel="license" href="/LICENSE.txt"/>
         <script type="module" src="/js/redirect.js"></script>
         <script type="module" src="/js/fastboot/dist/fastboot.min.mjs?1.0.8"></script>
-        <script type="module" src="/js/web-install.js"></script>
+        <script type="module" src="/js/web-install.js?1"></script>
```
thestinger: Should no longer add the cache busting strings. It's handled automatically now for `/js/*.js` but not the fastboot library.

static/js/web-install.js
```diff
@@ -155,12 +157,14 @@ async function downloadRelease(setProgress) {
...
 
     // Download and cache the zip as a blob
+    safeToLeave = false;
     setProgress(`Downloading ${latestZip}...`);
     await blobStore.init();
     await blobStore.download(`${RELEASES_URL}/${latestZip}`, (progress) => {
         setProgress(`Downloading ${latestZip}...`, progress);
     });
     setProgress(`Downloaded ${latestZip} release.`, 1.0);
+    safeToLeave = true;
```
thestinger: It also needs to handle failure.

static/js/web-install.js
```diff
@@ -299,4 +305,12 @@ if ("usb" in navigator) {
     console.log("WebUSB unavailable");
 }
 
+// This will create an alert box to stop the user from leaving the page
+window.onbeforeunload=function() {
```
thestinger: Should have spaces around `=`.

static/js/web-install.js
```diff
@@ -7,6 +7,8 @@ const RELEASES_URL = "https://releases.grapheneos.org";
 const CACHE_DB_NAME = "BlobStore";
 const CACHE_DB_VERSION = 1;
 
+var safeToLeave = true;
```
thestinger: This should be `let` instead of `var`. No `var` should be used in our code. It's a legacy way of declaring variables.

thestinger: Ideally this would be checked by `eslint`. It might be possible.

static/js/web-install.js
flawedworld: Solved

static/js/web-install.js
flawedworld: Solved

static/js/web-install.js
flawedworld: Opening additional pull for eslint changes. Also resolved.

flawedworld: See: #202

static/js/web-install.js
```diff
@@ -299,4 +306,12 @@ if ("usb" in navigator) {
     console.log("WebUSB unavailable");
 }
 
+// This will create an alert box to stop the user from leaving the page during actions
+window.addEventListener("beforeunload", stopUser => {
```
thestinger: Prefer naming `stopUser` as `event` by convention since it's the event being handled.

static/js/web-install.js
```diff
@@ -299,4 +306,12 @@ if ("usb" in navigator) {
...
+    if (safeToLeave == false) {
```
thestinger: Should be `!safeToLeave` instead of `safeToLeave == false`.

static/js/web-install.js
```diff
@@ -270,6 +276,7 @@ function addButtonHook(id, callback) {
             }
         } catch (error) {
             statusCallback(`Error: ${error.message}`);
+            safeToLeave = true;
```
thestinger: Probably best to drop this and do it a cleaner way.

static/js/web-install.js
```diff
@@ -196,13 +200,15 @@ async function flashRelease(setProgress) {
...
     setProgress("Flashing release...");
+    safeToLeave = false;
     await device.flashFactoryZip(blob, true, reconnectCallback,
         (action, item, progress) => {
             let userAction = fastboot.USER_ACTION_MAP[action];
             let userItem = item === "avb_custom_key" ? "verified boot key" : item;
             setProgress(`${userAction} ${userItem}...`, progress);
         }
     );
+    safeToLeave = true;
```
thestinger: I think the best way to do this is with `try {} finally { safeToLeave = true; }` to make sure it always runs and in the smallest scope possible right after setting it to `false`.

iNorman77:
Hi All,Does anyone here know how to enable tethering on pixel 2 using  graphene, need help please. Thanks 2:00 am, 2 April 2021, Daniel Micay ***@***.***>:
@thestinger commented on this pull request.



In static/js/web-install.js:
>      await device.flashFactoryZip(blob, true, reconnectCallback,
         (action, item, progress) => {
             let userAction = fastboot.USER_ACTION_MAP[action];
             let userItem = item === "avb_custom_key" ? "verified boot key" : item;
             setProgress(`${userAction} ${userItem}...`, progress);
         }
     );
+    safeToLeave = true;

I think the best way to do this is with try {} finally { safeToLeave = true; } to make sure it always runs and in the smallest scope possible right after setting it to false.

—You are receiving this because you are subscribed to this thread.Reply to this email directly, view it on GitHub, or unsubscribe.
s0ftC3£L

static/js/web-install.js
flawedworld: Sorted.

static/js/web-install.js
```diff
@@ -196,13 +206,21 @@ async function flashRelease(setProgress) {
...
+    safeToLeave = false;
+    try {
+        await device.flashFactoryZip(blob, true, reconnectCallback,
+            (action, item, progress) => {
+                let userAction = fastboot.USER_ACTION_MAP[action];
+                let userItem = item === "avb_custom_key" ? "verified boot key" : item;
+                setProgress(`${userAction} ${userItem}...`, progress);
+            }
+        );
+    } catch (error) {
```
thestinger: Can you leave out this catch block? It shouldn't be necessary.

static/js/web-install.js
```diff
@@ -155,11 +157,19 @@ async function downloadRelease(setProgress) {
...
     setProgress(`Downloading ${latestZip}...`);
     await blobStore.init();
-    await blobStore.download(`${RELEASES_URL}/${latestZip}`, (progress) => {
-        setProgress(`Downloading ${latestZip}...`, progress);
-    });
+    try {
+        await blobStore.download(`${RELEASES_URL}/${latestZip}`, (progress) => {
+            setProgress(`Downloading ${latestZip}...`, progress);
+        });
+    } catch (error) {
```
thestinger: Can you leave out this catch block? It shouldn't be necessary.

PR #208, opened 2021-04-12
Add emphasis on volume down.

Peter-Easton:


thestinger:
I don't really like this wording. I would like to improve it but I don't think this is what I want to do.

PR #210, opened 2021-04-15
Add to usage guide about carrier requesting disabling LTE only mode.

Peter-Easton:


flawedworld:
This is now no longer the case as we have patched out those checks.

PR #211, opened 2021-04-17
Add FAQ entry about containing invasive apps.

Peter-Easton:


thestinger:
Not planning to include this as is and it'll need substantial changes.

Peter-Easton:
Sure, I got ya. I'll go and take them up with you when you have some free time.

PR #218, opened 2021-05-04
Add support for changing release channel for WebUSB installation.

flawedworld:


thestinger:
I'm not sure if I want to do this. I don't want to cause confusion by having people think they need to install from the Beta channel in order to use the Beta channel.

flawedworld:
Perhaps to mitigate against the risk of this, the user should only see this dropdown if they enable a checkbox declaring they understand how beta builds work and that they can switch channels. An "Advanced users only" checkbox perhaps?

thestinger:
Not currently planned. I don't see a point in the added complexity and think it will contribute to people being more confused about release channels than they already are.

PR #229, opened 2021-06-16
Grammarly suggestions

khalykbek:
The word "set up" seems to be miswritten. Consider replacing it.


thestinger:
I think the existing way is correct since this is being used as a verb to refer to it being set up already.

PR #230, opened 2021-06-16
Grammarly suggestions

khalykbek:
Punctuation in compound/complex sentences


thestinger:
I think this hurts rather than helping clarity of the sentence.

PR #236, opened 2021-06-27
Use security headers for Matrix well-known server

ghost:
Tested these on my Matrix homeserver with no issues.

### DIFF
--- a/nginx/nginx.conf
+++ b/nginx/nginx.conf
@@ -232,6 +232,9 @@ http {
         }
 
         location = /.well-known/matrix/server {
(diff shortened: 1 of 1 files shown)


thestinger:
You don't need to set `Access-Control-Allow-Origin: *` for this. It's only needed for the client server. It's more of an anti-security header than a security header since it needs to be accessed cross-origin by web clients.

thestinger:
The other headers don't need to be set explicitly here because if you have no `add_header` directives, it's inherited from the context above.

PR #239, opened 2021-07-08
Fix order of table of contents

frkt3m:
Accessibility section comes first before System Navigation, but it is the other way around in the Table of Contents


thestinger:
I think it makes sense to start with system navigation since it's the most basic thing to know so I reordered the sections rather than fixing the table of contents.

PR #248, opened 2021-09-01
Added Markdown Document with 'howto' on google play extended directions

munntjlx:
Added a markdown document (the HTML version looked horrible). this is the extended docs that I was talking about last night.


thestinger:
These instructions have been improved.

PR #250, opened 2021-09-05
Support Pixel 5a in web installer

flawedworld:


thestinger:
Going to use the existing change that I made as part of adding it to the documentation and releases page. I just wasn't quite ready to deploy it yet since the current release needed a rebuild.

PR #252, opened 2021-09-11
index.html receives some fairly pedantic grammar fixes.

varikvalefor:


thestinger:
This needs to be split up into something that be considered / reviewed in a manageable way. It's not possible to do it this way since everything is bundled together and it's too hard to take the portion we want. Smaller pull requests are always better.

PR #269, opened 2021-10-21
Add mention of Android 12 clipboard warning to FAQ

Peter-Easton:


thestinger:
@Peter-Easton You can remove "an app in the foreground" because it's a given that only the focused app and clipboard manager (keyboard) can access the clipboard. You should also explain that the notification is only displayed when an app reads data in the clipboard that was written by another app. It's not shown when an app reads data from the clipboard that it put there itself, since that would just be an annoyance.

Peter-Easton:
Sure. I've changed it to this, let me know if I should squash it. 

**[superceded, see below]**

Peter-Easton:
Sorry, decided to give it one more minor edit, since I realized that it was ambiguous. 

```
<p>Android 12 further builds on this by displaying a toaster notification whenever an
app accesses clipboard data that was copied from a different app. This notification is
on by default and can be toggled under Settings ➔ Privacy ➔ Show clipboard access. When
on, it is not shown when an app copies and pastes content it put there itself to avoid
annoying the user.</p>
```

ghost:
Oh by the way, it's "toast" notification not toaster. :P

thestinger:
Incorporated this as https://github.com/GrapheneOS/grapheneos.org/commit/6d6a0422c4ef8d0e61e4eebea539d76fd4b1ba36.

Issue #275, opened 2021-11-01
Change the "Sandboxed Play Services" h2 on the usage page to something like "Play Services Compatibility Layer" 


TempUser13:
People seem to get confused about the phrasing "Sandboxed Play Services"

thestinger:
There aren't currently plans to rename this from Sandboxed Google Play.

PR #276, opened 2021-11-02
Update the "Sandboxed Play Services" h2 in usage.html

TempUser13:
Change the h2 for Sandboxed Play Services
Change a few words here and there

For issue #275


thestinger:
The current name is what we want to have, and if the id is changed then a redirect needs to be made.

PR #295, opened 2021-11-19
Add filesystem hardening details on the features pa

Lelmister101:
… ge. This PR adds a bit more detail on filesystem access hardening and the mitigations that are enabled.


Lelmister101:
This PR is a bit screwed up since I first made a bad mistake, and then GitHub thought that fetching and merging commits meant that it had to add a *new* commit authored by me (I likely shouldn’t have done that). It’s best just to take the text instead of merging this. Text:

Filesystem access hardening:
- Hardened mount permissions
- Link race mitigations enabled
- File spoofing mitigations enabled

PR #303, opened 2021-11-29
Improve index

ghost:


Shepherd115:
Looks really nice

thestinger:
Needs to be split up into multiple pull requests.

thestinger:
I want to keep the padding, line spacing, header sizes, etc. consistent across all the pages. That should be done in a separate pull request focused on those things so it can be reviewed / considered separately instead of bundling it all together. The best way to approach this is making small incremental improvements split up into independent pieces to the extent possible.

PR #306, opened 2021-12-04
replace obsolete block-all-mixed-content

ghost:


thestinger:
I think the correct approach is removing it rather than replacing it. We don't need requests to be upgraded to secure since we always use https.

thestinger:
According to the standard, it appears that `block-all-mixed-content` is essentially the default behavior and mixed content is required to be blocked or upgraded.

thestinger:
Going to keep it temporarily for backwards compatibility since not everything has adopted the new mixed content rules. The way they right documentation on MDN is not really accurate and is often misguided.

PR #313, opened 2021-12-11
Capitalize title text after colon

ghost:


thestinger:
I prefer it the way it was especially since we're not using title case.

PR #314, opened 2021-12-11
Add explicit width for phone img

ghost:


thestinger:
This seems wrong since it's not exactly 270. It would be better to redo this with an image size that can be more cleanly divided. It's not a high priority at all but it would be nice to get this done because it's a site rendering performance issue right now.

PR #322, opened 2021-12-19
Details to "Auditor" on usage.html

Shepherd115:
Dont know is this a thing that is wanted. Just thought it would be nice to have some information in here instead of a link.


Shepherd115:
Can just close this pr if its nothing that is wanted btw

thestinger:
I think it's better to leave this to the attestation site for the time being. It can be filled in with a summary later after improving that.

PR #323, opened 2021-12-20
Erase FIPS partition during installation

flawedworld:
was accidentally removed from release notes


thestinger:
I realized this shouldn't be included since this will be the first mainline release with Pixel 6 / Pixel 6 Pro support.

PR #328, opened 2021-12-29
Add bootflow details to install pages

Shepherd115:
Many people in the Matrix rooms seems to be confused about the "Your device is loading a different OS" message and the Google logo so thats why I think this would be a good idea.


thestinger:
This could still be added but I haven't had time to consider it properly.

PR #333, opened 2022-01-14
Add information on MediaDRM identifier

flawedworld:


static/faq.html
```diff
@@ -519,6 +519,11 @@ <h3><a href="#hardware-identifiers">Can apps access hardware identifiers?</a></h
                     around for compatibility. It used to need more extensive changes such as disallowing
                     access to the serial number but those restrictions are now standard.</p>
 
+                    <p>GrapheneOS adds the MediaDRM identifier permission which is revoked by default for
```
thestinger: It makes more sense to say disabled by default.

thestinger: This should probably also be turned into a usage guide section explaining that users may need to enable this to get DRM media working in an app.

PR #348, opened 2022-01-20
Update path to updater settings + document updater notification settings

ghost:


static/usage.html
```diff
@@ -314,6 +313,10 @@ <h3><a href="#updates-settings">Settings</a></h3>
                     device after an update once it has been idle for a long time. When this setting is
                     enabled, a device can take care of any number of updates completely automatically even
                     if it's left completely idle.</p>
+
+                    <p>The "Notification settings" option takes you to the System Updater notification
+                    settings which allows you to control the types of notifications you want to see from
```
thestinger: It allows more than controlling the types of notifications. It should also probably be made clear that it's simply a shortcut to the usual settings.

ghost: Clarified it's a shortcut and elaborate more on what you can control

PR #349, opened 2022-01-22
Update settings path to Android 12 disabling time sync and document new toggle disabling NITZ

ghost:
Since Android 12, the original setting "Use network-provided time" has been changed to "Set time automatically" and NITZ can be disabled with the new option "Set time zone automatically".


thestinger:
Note disabling automatic time zone isn't new but rather we just didn't document it before.

PR #362, opened 2022-01-27
Add app link verification to list of default connections made by GrapheneOS

ghost:
Linked to the usage guide section on how the user can disable it if they choose to.

### DIFF
--- a/static/faq.html
+++ b/static/faq.html
@@ -799,6 +799,10 @@
                         <li>
                             <p>DNS resolution for other connections</p>
                         </li>
+                        <li>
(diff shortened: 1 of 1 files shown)


thestinger:
It's a special case since as far as I know we don't have any app links in the base OS now that we removed the unintended wallpaper one. It's not really a default connection but rather a connection users enable when they install apps with app links. It's probably worth mentioning underneath the list but I don't think it makes sense to list it as a default connection.

ghost:
I can elaborate on it so it's an app that provides app link association checking which is normal when you start to install more and more apps.

thestinger:
I don't really think this should be included in the main list. It could be a section afterwards.

PR #365, opened 2022-01-31
install/cli: remove signify silent option, since an verified output is expected

brandsimon:
On ArchLinux there is no output at all, which is very confusing. Removing the -q option gives an output:
```
Signature Verified
oriole-factory-2022011423.zip: OK
```

### DIFF
--- a/static/install/cli.html
+++ b/static/install/cli.html
@@ -410,8 +410,9 @@ curl -O https://releases.grapheneos.org/DEVICE_NAME-factory-2021110122.zip.sig</
 
                 <pre>signify -Cqp factory.pub -x DEVICE_NAME-factory-2021110122.zip.sig &amp;&amp; echo verified</pre>
 
-                <p>This will output <code>verified</code> if verification is successful. If something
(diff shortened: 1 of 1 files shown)


thestinger:
The reason we disabled the output is because it will say `Signature Verified` even if the file verification fails because it's performing verification of the file separately. We were worried that an attacker would be able to trick users by providing the official signature with a tampered release so that the signature verifies but the file fails verification. It's unfortunate that signify doesn't use pre-hashing but rather directly uses ed25519 since it means memory usage scales up based on the size of the file and it gets very slow too. They support signed BSD hash files which is what we're using to work around it. minisign has pre-hashing via an -H option which is more convenient than this. It would be nice if signify got extended to support that.

brandsimon:
@thestinger 
Thank you very much for the clarification. It sounds very reasonable to use the quiet mode then.
I updated to PR and changed the text to include the empty output case.

thestinger:
The command has `&& echo verified` for the purpose of producing output on success though. I think you must have left that part out?

brandsimon:
@thestinger 
Yes, sorry, I have not copied to complete line :(

PR #439, opened 2022-05-19
This is now handled by backporting upstream patch

quh4gko8:
Recently, Chromium dev team has updated public gms library here, https://source.chromium.org/chromium/chromium/src/+/b45469c1f4f4cdaa14813db8d9fbe455fabd8459

It can be backported and cherry-picked in the future versions until it lands on a major version tag of Chromium


quh4gko8:
Done at https://github.com/GrapheneOS/grapheneos.org/commit/aec0a72d7f93e5b0d396ff6d1321680bf3a8928f

PR #464, opened 2022-07-04
Further clarify extended support releases

flawedworld:


thestinger:
I would prefer to word this differently.

PR #467, opened 2022-07-09
Adding DSDS hint to eSIM section of usage guide.

jonathancross:
Adding info requested here: https://github.com/GrapheneOS/os-issue-tracker/issues/1196#issuecomment-1163011728

Also confirmed it is correct and solves the problem on pixel 4a.

### DIFF
--- a/static/usage.html
+++ b/static/usage.html
@@ -999,6 +999,11 @@
 
                     <p>By enabling the toggle, the proprietary Google functionality is enabled and
                     will be used by the OS to provision and manage eSIMs.</p>
+
+                    <p>Note that if the eSIM installation process does not progress past the
+                    "Checking network info..." stage despite having a stable Internet connection,
+                    you may need to call the USSD code <code>*#*#4636#*#*</code> and then enable
+                    DSDS in the menu that is presented.</p>
                 </section>
             </section>
 


static/usage.html
```diff
@@ -999,6 +999,8 @@ <h3><a href="#sandboxed-google-play-esim">Privileged eSIM management</a></h3>
 
                     <p>By enabling the toggle, the proprietary Google functionality is enabled and
                     will be used by the OS to provision and manage eSIMs.</p>
+
+                    <p>NOTE: you may need to <a href="https://github.com/GrapheneOS/os-issue-tracker/issues/1196#issuecomment-1127466518">enable DSDS</a> if you find that you the eSIM installation process does not progress past "Checking network info..." stage.</p>
```
ghost: 1. Avoid using all caps words unless they're technical terms for better readability. Use "Note you may [..]"
2. Please split all this up into individual line breaks so it's easier for editing the HTML. See how other paragraphs look in the source code for maximum line width.

ghost: Also the code the user posted is not correct. It's `*#*#4636#*#*`

jonathancross: Thanks.  I added the USSD code directly to this usage guide for clarity.

flawedworld:
good to merge

PR #477, opened 2022-07-30
Add "Origin-Agent-Cluster" header

Seirdy:
Hint to browsers that we prefer per-origin process isolation. This
disables certain unsafe features regarding cross-origin same-site
resource sharing.

https://web.dev/origin-agent-cluster/#limitations

Specification link:
https://html.spec.whatwg.org/multipage/origin.html#origin-keyed-agent-clusters

This is just a hint to browsers. Depending on resource availability,
they may or may not actually allocate a process. For this reason, it's
not a robust security feature although it is preferable.

This header needs to be active on all pages from an origin for it to
work.


thestinger:
@Seirdy I'd rather omit the comments since it's the standard use case for the header and it's self-explanatory based on the definition of the headers.

thestinger:
Deployed this for our other sites too.

PR #479, opened 2022-08-06
CSP: block WebRTC

Seirdy:
Merged to the webappsec-csp repo in April:
https://github.com/w3c/webappsec-csp/pull/457

### DIFF
--- a/nginx/nginx.conf
+++ b/nginx/nginx.conf
@@ -376,7 +376,7 @@ http {
 
         location = /install/web {
             include snippets/security-headers-base.conf;
-            add_header Content-Security-Policy "default-src 'none'; child-src 'self'; connect-src 'self' https://releases.grapheneos.org/; font-src 'self'; img-src 'self'; manifest-src 'self'; script-src 'self'; style-src 'self'; form-action 'none'; frame-ancestors 'none'; block-all-mixed-content; base-uri 'none'" always;
+            add_header Content-Security-Policy "default-src 'none'; child-src 'self'; connect-src 'self' https://releases.grapheneos.org/; font-src 'self'; img-src 'self'; manifest-src 'self'; script-src 'self'; style-src 'self'; form-action 'none'; frame-ancestors 'none'; block-all-mixed-content; base-uri 'none'; webrtc 'block'" always;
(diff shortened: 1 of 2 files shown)


thestinger:
Is it implemented by any browsers yet?

Seirdy:
Not yet, unfortunately. I expect Chromium will implement this.

I think it'll be an important directive since WebRTC does not go through Fetch. Closing this loophole was one of the most demanded CSP features in the past.

Seirdy:
Upon further inspection: Permissions-Policy can already block WebRTC. 
I'll ask some webappsec folks about the logic behind adding the CSP 
directive in light of Permissions-Policy and report back.

Seirdy:
It looks like this is unnecessary, given that the Permissions-Policy already blocks WebRTC.

thestinger:
I don't think there's a Permissions-Policy setting for WebRTC itself.

ghost:
https://github.com/w3c/webappsec-permissions-policy/issues/250

This issue seems to indicate that there is no webrtc Permissions-Policy

Seirdy:
[@]dveditz:mozilla.org shared some info from the `#security:mozilla.org` Matrix room:

> found the difference -- Permission-policy controls things that request permissions, so it covers the camera and microphone getUserMedia() requests used in WebRTC. That doesn't prevent the connections though so that's why people want to add webrtc to CSP. That's a fairly recent addition (it's in the editor's draft, but not in the year-old published working draft) and may or may not be the best place to restrict it.
> Some of the https:// fetches (to get peer info from an ICE server, for example) might get blocked by connect-src, though it's not very clear to me which circumstances do or don't in the Firefox implementation at a fairly quick look. The DTLS communication is definitely not covered by fetch-centric CSP without something like this new proposal.

Direct link: https://matrix.to/#/!anEyEXBYVSlveMBsbt:mozilla.org/$8cPpnG-rvikEOwSTRVY-v_C-msCIQKaDZY2egDRy86Q?via=mozilla.org&via=matrix.org&via=kde.org

Honestly, I think that simply blocking JS from the subset of pages that don't need it would be a much more robust solution in the meantime.

thestinger:
@Seirdy We had to rebase the repository to fix some commit messages for a legal reason. Can you rebase this?

Seirdy:
On Fri, Aug 12, 2022 at 07:34:15AM -0700, terezipyrope wrote:
>https://github.com/w3c/webappsec-permissions-policy/issues/250
>
>This issue seems to indicate that there is no webrtc Permissions-Policy

Ah, I missed that. A full list of all implemented PP directives are in https://chromedevtools.github.io/devtools-protocol/tot/Page/#type-PermissionsPolicyFeature.

(side-note: hot damn this header is massive...almost 1.4kb just to opt-out of everything).

WofWca:
Gecko ticket: https://bugzilla.mozilla.org/show_bug.cgi?id=1783489
Chromium: https://bugs.chromium.org/p/chromium/issues/detail?id=1225968&q=webrtc%20csp&can=2

thestinger:
This is implemented now.

PR #503, opened 2022-08-29
doc: auto get number of cpu core

ghost:
not all machines have 16 cores,
to avoid any error we can automatically recover the number of cores with nproc

-j16 -> -j$(nproc)


thestinger:
repo sync jobs isn't really about the number of cores. If you have 8 cores, there's still a benefit to using 32 jobs. The limit on the number of jobs is their frustratingly strict rate limiting.

ghost:
thank you for your clarification

PR #504, opened 2022-08-30
Feat: use TLS record padding

Seirdy:
Use TLS record padding to make records a multiple of 1024 bytes,
impeding traffic analysis.


thestinger:
Something wrong with the whitespace.

Seirdy:
On Tue, Aug 30, 2022 at 12:15:22PM -0700, Daniel Micay wrote:
>Something wrong with the whitespace.

Fixed.

-- 
Seirdy (https://seirdy.one)

thestinger:
We can keep this in mind for the future but I don't currently see much point in doing it for our website where TLS is crucial for authentication but the encryption isn't really hiding much since it's obvious someone is accessing our website to read content about GrapheneOS and there isn't a diverse set of content to browse. It would be more valuable for the discussion forum or Matrix where sizes of messages could potentially leak sensitive information rather than just giving hints about which pages on our site are being read. I'm on the fence about it for those kinds of services, but I don't really want to do it for the website.

PR #508, opened 2022-09-07
Update certbot-ocsp-fetcher

Seirdy:
certbot-ocsp-fetcher now can use a Systemd cache dir and service sandboxing. See tomwassenberg/certbot-ocsp-fetcher#19

Depends on https://github.com/GrapheneOS/infrastructure/pull/10


thestinger:
This is implemented now.

PR #553, opened 2022-11-17
cancel snapshot update if one is in progress

maade93791:
Signed-off-by: randomhydrosol <randomhydrosol@glassrom.org>


thestinger:
Needs to be added for the CLI flashing approach to keep them in sync.

PR #576, opened 2022-12-29
Cancel snapshots on devices which were capable of doing so on all boo…

flawedworld:
…tloader versions


flawedworld:
https://github.com/GrapheneOS/grapheneos.org/pull/577

PR #577, opened 2022-12-29
Cancel snapshots on devices which are always capable of doing so

flawedworld:


static/js/web-install.js
```diff
@@ -148,6 +148,8 @@ const legacyQualcommDevices = ["sunfish", "coral", "flame", "bonito", "sargo", "
 
 const tensorDevices = ["cheetah", "panther", "bluejay", "raven", "oriole"];
 
+const day1SnapshotCancelDevices = ["barbet", "redfin", "bramble", "cheetah", "panther", "bluejay", "raven", "oriole"];
```
thestinger: Can you order this from newest to oldest like the others?

flawedworld: done

PR #582, opened 2023-01-09
Small features page updates

x86pup:
Add some various misc features, update the last written GKI version (stock OS is still the same), and change the link of Seedvault to GrapheneOS's fork as it isn't 100% accurate to link to the upstream project.


thestinger:
@r3g-5z GKI commit needs to be dropped since I did that part myself after doing the GKI update.

x86pup:
@thestinger Done

PR #590, opened 2023-01-31
 Add an overflow-x:auto; for the table

c2tz:


thestinger:
@1c2t1z This needs to use the standalone CSS file since inline CSS isn't permitted by the Content Security Policy.

c2tz:
@thestinger ouch I hadn't thought about that. No way to add a hash `'sha256-+/8l6fxQyrdundiK66ViWS5VlHiQUAi4V/Ir8XXFYh0='` in the csp with the integrity of the <style>? If not, I'll see if I can make a media query (but it will change the look of the array) it will be a bit more complicated as you said at the beginning. Because this method with the div can only work inline.

x86pup:
@1c2t1z It would require allowing unsafe-hashes which is still not wanted.

thestinger:
@1c2t1z You should be able to just give the div an id but I don't really understand why a div is required at all.

c2tz:
I tried based on this [example](https://www.w3schools.com/howto/tryit.asp?filename=tryhow_css_table_responsive) to add the overflow in the table class but it does not give the expected result

c2tz:
ok i found it in fact only with a div associated with an overflow that we can have the scroll bar (i don't know why) i will make the modifications and redo my pull request properly.

PR #591, opened 2023-02-01
Clarify and link to Google factory images for adevtool

x86pup:
It's currently non-obvious where the user has to obtain at least the BUILD_ID

and fix minor formatting consistency error (see formatting from other uses of `<pre>`)

Wasn't sure if I should use `factory image` or ` stock OS`, can change if necessary.

### DIFF
--- a/static/build.html
+++ b/static/build.html
@@ -510,11 +510,12 @@ repo sync -j16</pre>
 source script/envsetup.sh
 m aapt2</pre>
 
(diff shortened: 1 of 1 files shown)


thestinger:
The BUILD_ID should be obtained from the GrapheneOS version rather than guessing which one is needed from there. We always set our BUILD_ID to the stock Pixel OS BUILD_ID instead of the AOSP one.

x86pup:
Ah okay wasn't aware of that.

PR #609, opened 2023-03-06
Update changelog to reflect new SUPL toggle

matchboxbananasynergy:
This PR retroactively changes a previous changelog where the IMSI / Phone number patch was applied to say that SUPL can be disabled in Settings > Location as it is preferable to the APN configuration and it works on all devices (unlike APN config, which doesn't work on Tensor Pixels).

I find myself quoting that change when people ask, so it would be beneficial to have up-to-date information on how SUPL can be disabled there.

### DIFF
--- a/static/releases.html
+++ b/static/releases.html
@@ -886,7 +886,7 @@
                     <p>Changes since the 2023011000 release:</p>
 
                     <ul>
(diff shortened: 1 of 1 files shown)


thestinger:
I'd rather not change it in a way that wasn't accurate then. Removed the note about APN configuration though.

PR #622, opened 2023-03-22
reorder GrapheneOS services privacy policy + add few missing ones

x86pup:
Broadcom PSDS, attestation key remote provisioning, DNS over TLS check

left out connectivitycheck.grapheneos.network as grapheneos.network seems dedicated to network stuff anyways

### DIFF
--- a/static/faq.html
+++ b/static/faq.html
@@ -1101,15 +1101,17 @@
                         <li>grapheneos.network</li>
                         <li>grapheneos.online</li>
                         <li>grapheneos.org</li>
-                        <li>releases.grapheneos.org</li>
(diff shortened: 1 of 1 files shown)


thestinger:
@r3g-5z dnscheck.grapheneos.org is not a service itself.

x86pup:
@thestinger It still goes through our servers, like Broadcom PSDS and SUPL, no?

thestinger:
It doesn't actually exist. It's a name that's looked up via your DNS resolver, which queries our authoritative DNS servers. It's not actually a service itself.

x86pup:
Removed it

PR #643, opened 2023-04-11
Document that updating disabled apps no longer enables them

matchboxbananasynergy:
This PR adds information to https://grapheneos.org/features#user-installed-apps-can-be-disabled regarding the fact that GrapheneOS adjusts the default AOSP behavior of re-enabling disabled apps when you update them.

It felt relevant as an addition to that section instead of adding it to the "Other features" section.

If there are issues with wording, or this needs to be in another section, feel free to close.


thestinger:
This should be in a separate section. Better to add it to other features.

matchboxbananasynergy:
Closing and will replace with new PR adding this to "Other features".

PR #657, opened 2023-04-27
Explain how to add a search engine to Vanadium and set as default

kekkoudesu:
It isn't obvious that Vanadium supports adding search engines, which leads users to open issues asking for this feature: https://github.com/GrapheneOS/Vanadium/issues/254

A short note explaining how to add a search engine and set it as the default search engine will help new users setup their new browser with their choice of privacy-respecting search engine.

### DIFF
--- a/static/usage.html
+++ b/static/usage.html
@@ -601,6 +601,22 @@
                 sites from each other rather than only containing content as a whole. The sandbox
                 has been gradually improving on the desktop but it isn't happening for their
                 Android browser yet.</p>
+                
+                <p>Vanadium supports custom search engines through the <a
+                href="https://developer.mozilla.org/en-US/docs/Web/OpenSearch">OpenSearch</a> format.
+                To add a custom search engine and set it as the default search engine, the process is 
+                as follows:</p>
+
+                <ol> 
+                    <li>Go to the website and perform a search.</li> 
+                    <li>Go to <strong>Settings > Search Engine</strong> and set the search engine 
+                    which now appears under <strong>Recently Visited</strong> as the default search 
+                    engine.</li> 
+                </ol>
+
+                <p>If the search engine does not appear even after performing a search, it likely 
+                does not serve an OpenSearch document, and cannot be added as a custom search 
+                engine.</p>
             </section>
 
             <section id="camera">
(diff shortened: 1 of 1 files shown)


jahway603:
@thestinger I'm following the process you've outlined in this PR and it does not work with Startpage.com, but I also followed this same process in Bromite with Startpage.com and it did work. Should I open a new issue?

thestinger:
Please use Matrix or the discussion forum to ask questions and don't specifically direct questions towards me.

thestinger:
Search engines not being compatible with mobile Chromium is not Vanadium specific and not a bug. They choose to do things in a way that's not compatible including using POST instead of GET for searches.

kekkoudesu:
@jahway603 
> I'm following the process you've outlined in this PR and it does not work with Startpage.com, but I also followed this same process in Bromite with Startpage.com and it did work. Should I open a new issue?

I'm the author of this PR. As explained previously, the reason Chromium-based browsers on Android like Vanadium don't pick up the OpenSearch document served by Startpage is because they have chosen to use POST HTTP requests by default for privacy reasons (this way, it won't show your searches in your browsing history), which don't expose the query parameter used for making searches in the URL. You can change the HTTP Request Method to GET here: https://www.startpage.com/do/settings

Once you do that, it will pick up the search engine properly. Additionally, you will need to do this in normal browsing mode, not Incognito Mode. Any future searches with Startpage will be made through the GET method. To replicate the benefit of not storing searches in history for Chromium-based browsers on Android like Vanadium, the only way to do this is to always use Incognito mode or regularly clear your history manually.

It would be less confusing if Chromium-based browsers allowed you to to simply add the search engine when an OpenSearch document is served as Firefox on desktop allows you to do as soon as you land on the website. You may wish to open a feature request about this on bugs.chromium.org, because while these browsers are properly parsing the OpenSearch document in most circumstances, they apparently decide to ignore it if the resulting search uses a POST method.

Chromium-based browsers on Android may allow you to add search engines manually sometime this year, if all goes well, so adding search engines may be less confusing in the future: https://bugs.chromium.org/p/chromium/issues/detail?id=1440189

PR #662, opened 2023-05-07
Mention move of PSDS setting to Location Settings

matchboxbananasynergy:
This explains that the PSDS setting was moved from Settings ➔ Network & Internet to Settings ➔ Location.

Some people are confused about it, so it would be good to mention that:

https://discuss.grapheneos.org/d/4870-grapheneos-version-2023050500-released/8


thestinger:
Should be a separate changelog entry.

PR #685, opened 2023-06-16
CSS: Enable smooth scrolling

tedgravlin:
This enables smooth scrolling thru CSS. This should make navigating the site a little more smooth.

| Before  | After |
| :-------------: | :-------------: |
|   |


thestinger:
Not sure if we should do this. We need to think about it.

PR #686, opened 2023-06-17
Newer versions of Debian and Ubuntu do provide up-to-date fastboot package

tobhe:
I was just following the installer and noticed that on newer versions of Debian and Ubuntu there is an easier way to get an up-to-date version of fastboot.

Tested on Ubuntu 23.04.

Sources:
https://packages.ubuntu.com/kinetic/google-android-platform-tools-installer
https://packages.debian.org/bookworm/google-android-platform-tools-installer

### DIFF
--- a/static/install/cli.html
+++ b/static/install/cli.html
@@ -233,9 +233,15 @@
 
                 <pre>sudo pacman -S android-tools</pre>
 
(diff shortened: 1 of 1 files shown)


thestinger:
This is out-of-date already and won't have proper Android 14 support.

PR #688, opened 2023-06-18
Add a scroll-to-top button

tedgravlin:
This adds a scroll-to-top button to most pages of the site.

This change will make navigating the site less tedious. For example, with the current behavior, if you scroll far down on a long page it's tedious to have to manually scroll all the way back to the top (especially on mobile).
The scroll-to-top button fixes this issue by automatically scrolling to the top of the page when tapped. It uses the same styling as other buttons from the site.

Demonstration:

https://github.com/GrapheneOS/grapheneos.org/assets/123981212/54e4707e-33a3-428b-84d2-06600784d0d4

### DIFF
--- a/static/articles/attestation-compatibility-guide.html
+++ b/static/articles/attestation-compatibility-guide.html
@@ -28,6 +28,7 @@
         <link rel="manifest" href="/manifest.webmanifest"/>
         <link rel="license" href="/LICENSE.txt"/>
         <link rel="me" href="https://grapheneos.social/@GrapheneOS"/>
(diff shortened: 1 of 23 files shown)


tedgravlin:
After fixing the indentation errors, I need to change the behavior of the button to use an event listener rather than calling the scrollUp() function directly. This should fix the other issue.

Then I'll mark this PR ready for review again.

tedgravlin:
Well the automated tests certainly gave me a run for my money, but everything should be good to go.

PR #690, opened 2023-06-19
Update behavior of navigation bar hover

tedgravlin:
Instead of highlighting the background of the button, this changes the color of the text being hovered over. This brings the behavior more inline with how other sites work.

Before:

https://github.com/GrapheneOS/grapheneos.org/assets/123981212/e5c70a53-fd03-427c-957f-47dd0ef1568d

After:

https://github.com/GrapheneOS/grapheneos.org/assets/123981212/9208fbb0-448c-4672-a040-93feacf7f92d

### DIFF
--- a/static/main.css
+++ b/static/main.css
@@ -116,6 +116,10 @@ var {
     display: flex;
 }
 
+#site-menu ul li:hover {
+    background-color: #424242; /* grey800 */
+}
+
 #site-menu ul li a {
     color: #fff; /* 100% white */
     display: flex;
@@ -136,7 +140,6 @@ var {
 
 #site-menu ul li[aria-current] a {
     color: #64b5f6; /* blue300 */
-    font-weight: bold;
 }
 
 #site-menu img {
--- a/static/main.css
+++ b/static/main.css
@@ -120,14 +120,14 @@ var {
     color: #fff; /* 100% white */
     display: flex;
     text-decoration: none;
-    padding: 0.25em 0.5em;
-    margin: 0.75em 0.5em;
+    padding: 0.50em 0.75em;
+    margin: 0.50em 0.25em;
     align-items: center;
 }
 
 #site-menu ul li a:hover {
     background-color: #424242; /* grey800 */
-    border-radius: 20px;
+    border-radius: 25px;
 }
 
 #site-menu ul li a:visited {
--- a/static/main.css
+++ b/static/main.css
@@ -116,16 +116,16 @@ var {
     display: flex;
 }
 
+#site-menu ul li:hover {
+    background-color: #424242; /* grey800 */
+}
+
 #site-menu ul li a {
     color: #fff; /* 100% white */
     display: flex;
     text-decoration: none;
     padding: 1em;
     align-items: center;
-}
-
-#site-menu ul li a:hover {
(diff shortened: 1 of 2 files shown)


thestinger:
I'm not sure about this. For example, we're currently on GitHub which uses the same approach we do for tabs and I think it's more easily visible.

tedgravlin:
> I'm not sure about this. For example, we're currently on GitHub which uses the same approach we do for tabs and I think it's more easily visible.

How about something more like this?

https://github.com/GrapheneOS/grapheneos.org/assets/123981212/61fb774e-27dc-49f0-b26f-11924debf07a

thestinger:
I think it needs to be larger and not so close to the words.

thestinger:
Changing the structure, organization and content of the site is what's really needed. A massive amount of high quality writing is needed. I don't think what's wrong with the site is aesthetics but rather poor content organization, major lack of content and in particular lack of audience-appropriate content such as covering each feature properly in the features page, in the usage guide for end users in a non-overly-technical way and in the article section of the section giving detailed technical information.

Also, adding useful, non-fluff graphical content throughout the site **after considering how things should be organized and rewritten** would be highly valuable. I think not having that is a major factor in the site not being great.

tedgravlin:
> I think it needs to be larger and not so close to the words.

Is this any better?

https://github.com/GrapheneOS/grapheneos.org/assets/123981212/b99f68fd-8201-4d42-9bb1-737bb2bc6a6f

> Changing the structure, organization and content of the site is what's really needed. 

That's true, a more friendly approach to the overall structure would certainly help. I'll try to come up with some ideas for improvements.

> Also, adding useful, non-fluff graphical content throughout the site **after considering how things should be organized and rewritten** would be highly valuable. I think not having that is a major factor in the site not being great.

Agreed. I think a nice balance can be found where the site can be both visually appealing and highly informational. But getting a proper structural redesign done before worrying about the visual appeal is important.

thestinger:
@tedgravlin Yes, but can you test it across Chromium, Firefox and Safari including on mobile for all 3?

tedgravlin:
>  can you test it across Chromium, Firefox and Safari including on mobile for all 3?

I was only able to test it in Chromium and Firefox for desktop, and wasn't able to test anything on mobile.
Unfortunately, I also don't have access to anything that can run Safari.

However, I did test it using the mobile device view in the web development tools of both Firefox and Brave on desktop.

I would think this change should work universally across all browsers as its just padding and color changes.

thestinger:
Can you test a WebKit browser? That would be a decent substitute for Safari. Maybe the GNOME one.

tedgravlin:
> Can you test a WebKit browser?

I can confirm that everything works as expected on the WebKit browser that I tried.

static/main.css
```diff
@@ -116,18 +116,20 @@ var {
...
     text-decoration: none;
-    padding: 1em;
+    padding: 0.50em 0.75em;
+    margin: 0.50em 0.25em;
     align-items: center;
 }
 
+#site-menu ul li a:hover {
+    background-color: #424242; /* grey800 */
+    border-radius: 25px;
```
tedgravlin: Sounds good to me

PR #707, opened 2023-07-13
Use simple templating for static processing

ghost:


process-templates.py
```diff
@@ -0,0 +1,48 @@
...
+
+# Get all .html.jinja files with path. The path is important to copy the
+# rendered template to the desired directory structure later.
+template_file_list = []
+for dirpath, dirnames, filenames in os.walk("templates/"):
+    for filename in filenames:
+        if filename.endswith(".html.jinja"):
+            # We remove the "template/" part from the path. The other parts of
+            # the path are important for the reason mentioned above.
+            template_file_list += [
```
thestinger: This should use `append(value)` instead of `+= [value]`.

thestinger (changes requested):

PR #709, opened 2023-07-14
Highlight data deletion as red in installation guide

tomkoid:
Everything about wiping data should be definitely highlighted as it is really easy to miss when reading the installation guide.


thestinger:
I don't think we need to do this and it will be distracting.

PR #723, opened 2023-08-04
install/cli: Add hint to Debian/Ubuntu signify-openbsd command

r4co0n:
On Debian and Ubuntu systems, we need to install `signify-openbsd` to continue with the CLI installation variant, as described in the step "Obtaining signify". However, this is not reflected in the step "Obtaining factory images", where `signify-openbsd` should be called instead of `signify` for these environments.

This PR adds commands to run in Debian/Ubuntu when trying to verify a GrapheneOS signify signature.


thestinger:
The earlier step does `alias signify=signify-openbsd`.

r4co0n:
I also realized this just now and came here to close this. Sorry for the noise :)

P.S. I had signify-openbsd already installed so skipped the rest of this section, which I shouldn't have.

PR #727, opened 2023-08-09
feat: extend the use of page templates

pimlie:
This PR further implements the use of templates for the pages. All pages now extend a default layout and call a macro to generate the meta tags for seo. ~I've also added a minimal dev server so it's easier to work on the templates, see the readme for details~ moved to #731

I have compared the generated static html files as follows:
```bash
$ git checkout main
$ ./process-static
$ find static-tmp/ -iname '*.br' -exec rm {} \;
$ find static-tmp/ -iname '*.gz' -exec rm {} \;
$ mv static-tmp static-main
$ git checkout feat-extend-page-templates
$ ./process-static
$ find static-tmp/ -iname '*.br' -exec rm {} \;
$ find static-tmp/ -iname '*.gz' -exec rm {} \;
$ git diff --no-index --word-diff static-main static-tmp
```

After comparing the only changes to the html left should be:
- The order of meta tags is different (because `main.css` is added outside of the seo meta tags)
- Single quotes in titles are escaped
- I've moved the upper most `id` tag from the `<main>` tag to the `<h1>`, as the main tag is now part of the (parent) layout template

This might help with reviewing a bit, please note that Github says there are 52 changed files but its basically just half of that as github fails to detect moving the html files from `static/` to `/templates/pages` _after_ updating the templates to use the layout file. I.e. if you check the first commit in this PR https://github.com/GrapheneOS/grapheneos.org/pull/727/commits/e45d7b48fc6f1acbad1c188d4f991fe4ba6b9716 then it will list the html files as being renamed. And if you check https://github.com/GrapheneOS/grapheneos.org/pull/727/commits/fb718006d58e298d3a71931afb46619405d0d3c7 then you'll see those moved files have been updated to extend the layout template. But when git then checks this PR at once it apparently won't see them as the 'same' files any more unfortunately...


pimlie:
Closing this PR as apparently there is no interest into merging these improvements and there are too many conflicts now

PR #752, opened 2023-09-25
Qualify Verizon Devices as Unsupported Due to OEM Unlock

MateusAmin:
Verizon Pixels do not support "OEM Unlocking". Even if they are carrier unlocked. This prevents the user from installing GrapheneOS .

There are no known workarounds or cases were Verizon has lifted the lock.

### DIFF
--- a/static/faq.html
+++ b/static/faq.html
@@ -190,6 +190,8 @@
                     and Bionic along with easily added support for other environments. It can easily run
                     on non-Linux-based operating systems too, and supporting some like HardenedBSD is
                     planned but depends on contributors from those communities.</p>
+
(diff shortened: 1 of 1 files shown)


MateusAmin:
> Thank you
> I believe you can tell if a device will be oem unlockable based on its model number
> Verizon and T-Mobile devices dont support it, International unlocked and Japanese models do
> 

- jasmonate_danzel from Matrix chat

I am digging a bit more into it. Might be better whitelist than to blacklist. Listing the ones known to be OEM unlockable and how to identify them. Skipping any brands that are mixed across supported devices for simplicities sake.

matchboxbananasynergy:
We can close this in favor of https://github.com/GrapheneOS/grapheneos.org/commit/dcf92cd45c85732b25be25b50aaa1168af60cfe0 and improve the existing wording or make it more specific later, if deemed necessary.

PR #759, opened 2023-10-06
give advice when installing from another Android device

orazioedoardo:


thestinger:
will be, not might be

thestinger:
I think this needs to be moved to the right part of the paragraph before it talks about locking.

orazioedoardo:
I thought about that but added there since it's closer to where the "Reconnect device" button appears. When the user clicks "Flash release" they are likely to continue reading and figure out this advice and also about not touching it by the time the first reboot occurs.

Another option is adding after this text, before the "Flash release" button:

> The initial install will be performed by flashing the factory images. This will replace the existing OS installation and wipe all the existing data.

PR #768, opened 2023-10-14
update app name following official rebrand to Pixel Camera

Wonderfall:


thestinger:
Need to add redirect to redirect.js.

PR #770, opened 2023-10-16
Update supported device list and move Pixel 5 entry to EOL section.

akc3n:
The end of life for the Pixel 5, was yesterday, October 14, 2023.


thestinger:
We should wait until next month for confirmation. They've extended device lifetimes by several months before.

akc3n:
Understood. Thanks for clarifying.

thestinger:
Pixel 5a has the same SoC and it would be somewhat disappointing if they didn't extend it past what they promised.

PR #787, opened 2023-11-07
remove CLI installation section about deactivating drivers_autoprobe

pieqq:
Several people reported that setting `/sys/bus/usb/drivers_autoprobe` to 0 as recommended in the installation guide resulted in the following step (`fastboot flashing unlock`) not working.

The workaround, as explained on the forum[1], is to ignore this section. Therefore, this commit removes it from the installation guide.

[1] https://discuss.grapheneos.org/d/3419-fastboot-waiting-for-device

### DIFF
--- a/static/install/cli.html
+++ b/static/install/cli.html
@@ -319,18 +319,6 @@ Installed as /home/username/platform-tools/fastboot</pre>
                 <p>Connect the device to the computer. On Linux, you'll need to do this again if
                 you didn't have the udev rules set up when you connected it.</p>
 
-                <p>On Linux, GNOME has a bug causing compatibility issues with the installation
-                process. It wrongly detects the device in fastboot mode or fastbootd mode as being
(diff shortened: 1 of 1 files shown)


thestinger:
This section is in both the web and CLI guides, and it's not recommended. It's provided as a workaround for broken operating systems.

pieqq:
I had missed the section in the Web documentation, sorry!

Maybe instead of removing it entirely, a note could be added to let people try without setting it to `0` first?

thestinger:
That makes sense. It's only there as a possible workaround. I'm not sure if removing it makes sense. The issue it was added to work around may be fixed in distributions with up-to-date software but Debian stable exists.

pieqq:
By any chance, do you know what GNOME bug we are referring to? This way we can make it more specific what versions of GNOME users should run the workaround commands.

PR #793, opened 2023-11-14
Make buttons inherit font family

cosmicproc:
Form elements don't inherit fonts by default. This causes ```<button>``` elements to have a browser-chosen font unlike the ```<a>``` element that looks like a button in the home page.
Also, if desired, ```<input>```, ```<select>```, and ```<textarea>``` elements, which have the same behavior, can be added to the CSS rule as well.

**Home page button (actually an ```<a>``` element) (https://grapheneos.org/)**
image1
**Buttons in the web installer page (https://grapheneos.org/install/web)**
image2
**These are actually different fonts**
image3
image4

Reference: https://stackoverflow.com/questions/26140050/why-is-font-family-not-inherited-in-button-tags-automatically

### DIFF
--- a/static/main.css
+++ b/static/main.css
@@ -159,6 +159,10 @@ main {
     overflow-wrap: break-word;
 }
 
+button, input, select, textarea {
+    font-family: inherit;
+}
+
 button, .button, .button:visited {
     color: white;
     font-size: 0.9em;


thestinger:
> Also, if desired, <input>, <select>, and <textarea> elements, which have the same behavior, can be added to the CSS rule as well.

That makes sense, although we don't currently use any on grapheneos.org but we use some on attestation.app.

Can you please submit this for https://github.com/GrapheneOS/AttestationServer too? We try to keep them mostly in sync, although the attestation site is a bit behind in some areas.

cosmicproc:
Sure

thestinger:
Not clear why GitHub is formatting that whole reply as a quote.

```
> Also, if desired, <input>, <select>, and <textarea> elements, which have the same behavior, can be added to the CSS rule as well.

That makes sense, although we don't currently use any on grapheneos.org but we use some on attestation.app.

Can you please submit this for https://github.com/GrapheneOS/AttestationServer too? We try to keep them mostly in sync, although the attestation site is a bit behind in some areas.
```

thestinger:
@cosmicproc Can you make this consistent with the attestation.app one even though we don't use these elements yet? It will be forgotten when we do use them otherwise.

PR #796, opened 2023-11-27
Adjusted logo

maki3000:
Very unimportant PR...

I really would like to contribute, but since I'm already overwhelmed to write nice PHP, I thought I could do design.

Make the logo a bit more appealing with rounded corner and a slight gradient. If you like it, of course, I could also adjust the logos of the other repositories.

Thank you for making GrapheneOS happen!


thestinger:
We probably don't want to do this but we appreciate the effort.

PR #824, opened 2024-01-02
Warn eSIM users to keep PUKs handy to avoid device lockout (as reported in OS issue 2162)

de0u:
This warns eSIM users to obtain and memorize the PUK to avoid being locked out of the device at boot time, as reported by three users in [OS issue #2162](https://github.com/GrapheneOS/os-issue-tracker/issues/2162) and also as discussed in [forum thread 9854](https://discuss.grapheneos.org/d/9854-gos-phone-unusable-after-forgetting-esim-pin-due-to-google-esim-management/14).


de0u:
Oops!  I will re-submit.  Apologies for the noise!

PR #825, opened 2024-01-02
Warn eSIM users to keep PUKs handy to avoid device lockout (as reported in OS issue 2162)

de0u:
This warns eSIM users to obtain and memorize the PUK to avoid being locked out of the device at boot time, as reported by three users in https://github.com/GrapheneOS/os-issue-tracker/issues/2162 and also as discussed in [forum thread 9854](https://discuss.grapheneos.org/d/9854-gos-phone-unusable-after-forgetting-esim-pin-due-to-google-esim-management/14).

### DIFF
--- a/static/usage.html
+++ b/static/usage.html
@@ -1123,6 +1123,19 @@
                     "Checking network info..." stage despite having a stable Internet connection,
                     you may need to call the USSD code <code>*#*#4636#*#*</code> and then enable
                     DSDS in the menu that is presented.</p>
+
+                    <!-- https://github.com/GrapheneOS/os-issue-tracker/issues/2162 -->
+                    <p><strong>Warning:</strong> Do not enable a PIN for an eSIM without first
+                    memorizing the eSIM's
+                    <a href="https://en.wikipedia.org/wiki/Personal_unblocking_key">PUK</a>.
+                    If you do not know your eSIM's PUK, you may be forced to factory-reset
+                    your device (<strong>thus wiping your data</strong>) to reinstall the
+                    stock OS, as follows. GrapheneOS automatically disables the proprietary
+                    Google eSIM management code on every reboot. If an eSIM has locked itself
+                    because an incorrect PIN has been entered multiple times, it will not be
+                    possible to disable the eSIM on the boot-time SIM-unlock screen.  Without
+                    the PUK it will not be possible to unlock the device, access Settings,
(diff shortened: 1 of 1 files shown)


static/usage.html
```diff
@@ -1123,6 +1123,19 @@ <h3><a href="#sandboxed-google-play-esim">Privileged eSIM management</a></h3>
...
                     you may need to call the USSD code <code>*#*#4636#*#*</code> and then enable
                     DSDS in the menu that is presented.</p>
+
+                    <!-- https://github.com/GrapheneOS/os-issue-tracker/issues/2162 -->
+                    <p><strong>Warning:</strong> Do not enable a PIN for an eSIM without first
+                    memorizing the eSIM's
+                    <a href="https://en.wikipedia.org/wiki/Personal_unblocking_key">PUK</a>.
+                    If you do not know your eSIM's PUK, you may be forced to factory-reset
+                    your device (<strong>thus wiping your data</strong>) to reinstall the
+                    stock OS, as follows. GrapheneOS automatically disables the proprietary
```
matchboxbananasynergy, line 1133:
```suggestion
                    stock OS, as follows. GrapheneOS automatically disables the proprietary
```
What does "as follows" refer to here?

de0u: "As follows" refers to the remainder of the paragraph, i.e., the steps by which a user might end up unable to unlock the device:
* Device reboot triggers automatic disabling of Google's eSIM code
* Multiple PIN entries fail, causing the PIN to lock itself
* The "Disable eSIM" button on the SIM-unlock screen won't work
* So only the PUK will do.

It would be possible to turn the multiple sentences into a formatted list, but that seemed like it might be overkill -- especially since it seems the desired situation would be to at some point fix the "Disable eSIM" button, or at least to replace it with an "Ignore eSIM" button, or to have native eSIM management code.  Overall it seemed as if the eventual fate of this warning would be deletion.

matchboxbananasynergy:
This is likely no longer necessary due to the refactoring of eSIM activation here: https://github.com/GrapheneOS/platform_frameworks_base/pull/480

Changelog which briefly goes over the changes:
https://github.com/GrapheneOS/grapheneos.org/commit/4174ecb6f7ebf2d5c7cd7280f8b3f1962fa22331

The toggle will no longer be automatically disabled every reboot, so it makes sense for it to be enabled to avoid these situations. A message also appears when toggling this that covers this edge case.

de0u:
Great news!

PR #844, opened 2024-01-22
add initial Android Auto documentation

muhomorr:


static/usage.html
```diff
@@ -1168,6 +1169,49 @@ <h3><a href="#sandboxed-google-play-esim">Privileged eSIM management</a></h3>
...
+                <p>The permission toggles ask for a confirmation before turning on. The
+                confirmation popup explains what access each permission toggle provides.</p>
+
+                <p>By default, Android Auto is not granted any kind of privileged access. It's
+                treated the same way other apps are treated.</p>
+
+                <p>In order to work, Android Auto has to be granted baseline permissions for wired
+                or wireless Android Auto. Wired Android Auto requires far less access than wireless
+                Android Auto does. Baseline permissions are controlled by the "Allow permissions
+                for wired / wireless Android Auto" toggles.<p>
```
matchboxbananasynergy:
```suggestion
                for wired / wireless Android Auto" toggles. In some cases, enabling the toggle for wireless Android Auto is required even if you intend to use it wired. Therefore, if wired Android Auto isn't working with the wired Android Auto toggle enabled, enabling the wireless Android Auto enable is a good initial troubleshooting step.</p>
```

quh4gko8: The end element in this paragraph had a typo as well, `</p>` should be placed instead of `<p>` to pass the checks

matchboxbananasynergy: > The end element in this paragraph had a typo as well, `</p>` should be placed instead of `<p>` to pass the checks

Added that. Thanks!

PR #851, opened 2024-01-26
Update Android Studio (Fix file name)

topminipie:


thestinger:
Thanks, I missed that. I checked the hash but not with the command from that page.

PR #855, opened 2024-01-30
refactor eSIM section

### LINKED ISSUE #853: Re-write documentation on eSIM activation/management on GrapheneOS to reflect latest changes
Since https://grapheneos.org/releases#2024012600, eSIM activation/management no longer requires sandboxed Google Play, along with other improvements to it.

The current documentation (https://grapheneos.org/usage#sandboxed-google-play-esim) is out of date.

The following needs to happen:

- The eSIM activation/management usage guide section should be moved away from sandboxed Google Play as it is no longer relevant to it
- Name of toggle should be changed to reflect the new one
- It should be made clear that due to the nature of its integration along with isolating it from non-system apps, no data is shared to Google as part of it
- It should be documented that the toggle is no longer reset on a reboot
- (Optionally) it should be documented that if you have an eSIM with a PIN associated with it, you should keep the toggle enabled at all times so that you don't risk being locked out of your device if you forget said PIN and don't have access to the PUK.

dd-dreams:
Closes https://github.com/GrapheneOS/grapheneos.org/issues/853

### DIFF
--- a/nginx/nginx.conf
+++ b/nginx/nginx.conf
@@ -271,6 +271,10 @@ http {
             return 301 /faq#default-connections;
         }
 
+        location = /usage#sandboxed-google-play-esim {
+            return 301 /usage#esim-support;
+        }
+
         # redirect away from the old SVG favicon location
         location = /mask-icon.svg {
             return 301 /favicon.svg;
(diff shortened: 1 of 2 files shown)


matchboxbananasynergy:
This needs a redirect from the current link to the new link at the very least. I filed https://github.com/GrapheneOS/grapheneos.org/issues/853 for this, which includes a list of things that it would be nice for it to include.

dd-dreams:
You mean a redirect from https://grapheneos.org/usage#sandboxed-google-play-esim to the new section?

matchboxbananasynergy:
Correct, if I recall correctly, this is done here: https://github.com/GrapheneOS/grapheneos.org/blob/main/nginx/nginx.conf

dd-dreams:
Let me know what you think.

PR #856, opened 2024-01-31
fix redirect to eSIM support

orazioedoardo:
URI fragment is not sent to the server so Nginx can't redirect to the section.

### DIFF
--- a/nginx/nginx.conf
+++ b/nginx/nginx.conf
@@ -271,10 +271,6 @@ http {
             return 301 /faq#default-connections;
         }
 
-        location = /usage#sandboxed-google-play-esim {
-            return 301 /usage#esim-support;
-        }
-
         # redirect away from the old SVG favicon location
         location = /mask-icon.svg {
             return 301 /favicon.svg;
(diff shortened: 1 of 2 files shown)


dd-dreams:
I think it's because Daniel might have not restarted the nginx server.

dd-dreams:
So if javascript is not enabled it won't work.

orazioedoardo:
> I think it's because Daniel might have not restarted the nginx server.

Your latest changes were incorporated so they likely did but [wouldn't work anyways.](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/Identifying_resources_on_the_Web#fragment:~:text=It%20is%20worth%20noting%20that%20the%20part%20after%20the%20%23%2C%20also%20known%20as%20the%20fragment%20identifier%2C%20is%20never%20sent%20to%20the%20server%20with%20the%20request.)

thestinger:
Yes, these fragment redirects need to be done via JavaScript. There's an explanation in the redirect.js file.

PR #859, opened 2024-02-02
Document Widevine provisioning default connection

### LINKED ISSUE #848: Document Widevine provisioning in default connections FAQ section
This section should include Widevine provisioning and the different options (GrapheneOS proxy or Standard) that are offered, and where that can be toggled.

matchboxbananasynergy:
This PR documents the Widevine provisioning connection, along with explaining where you can change from the default reverse proxy to the standard Google connection.

This closes https://github.com/GrapheneOS/grapheneos.org/issues/848.

### DIFF
--- a/static/faq.html
+++ b/static/faq.html
@@ -1039,6 +1039,16 @@
                             own service and we'd need to support proxying to that service too.</p>
                         </li>
                         <li>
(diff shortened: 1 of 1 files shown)


matchboxbananasynergy:
Converting to draft as some things are unclear:

- When the connection happens (default connection, or just when relevant services are used)
- Needs to be explained what the connection does.

thestinger:
I think this needs to be moved to other connections since I think it's always triggered by apps using it and Vanadium has protected media off by default.

PR #879, opened 2024-02-21
List SetupWizard2 instead of the legacy SetupWizard repository

matchboxbananasynergy:


matchboxbananasynergy:
Unless the old SetupWizard should also be listed since it's used on legacy extended support devices?

PR #881, opened 2024-02-21
Overhaul community section; explicit recommendation of Discord; expand on descriptions of each platform along with pros/cons

matchboxbananasynergy:
This PR makes various changes to the community section on the contact page. Namely:

- Explicit recommendation of Discord over other supported platforms due to better moderation tools along with being able to prevent seeing disturbing content.
- Added more detail about state of each platform and how they are configured.

### DIFF
--- a/static/contact.html
+++ b/static/contact.html
@@ -66,8 +66,8 @@
                     <h3><a href="#community-chat">Chat</a></h3>
 
                     <p>GrapheneOS has a very active community primarily based around the official
-                    chat rooms. Our chat rooms are bridged across Discord, Telegram and Matrix so
-                    you can choose your preferred platform.</p>
+                    chat rooms. Our chat rooms are bridged across Discord (recommended), Telegram 
+                    and Matrix so you can choose your preferred platform.</p>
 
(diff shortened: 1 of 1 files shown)


thestinger:
Don't know if this is a good idea.

matchboxbananasynergy:
Discord is our fastest growing platform, the one in which the user experience is the best, along with being a platform which allows us to safeguard community members from witnessing disturbing content through raids. While it is true that the latter also applies to Telegram, it comes at a cost (e.g. Telegram users cannot send media while Discord users are able to do so due to Discord's native moderation tools preventing potential for abuse). That, along with reduced readability don't make it an ideal option, but still a decent option for people who already use Telegram and not Discord.

Matrix is near impossible to actually secure and we are not in a position to properly safeguard it so that it is a safe experience for community members, so it can also not be recommended right from the get go.

While I understand the benefits of providing all options and letting users choose without steering them in any which way, I think the benefits of Discord are so substantial from a UX and community safety standpoint that making the recommendation makes sense. Someone might not want to read all paragraphs explaining the details of each platform and their pros/cons. In such a case, I want them to know there is a reason for them prefer Discord.

matchboxbananasynergy:
Will revisit.

PR #882, opened 2024-02-22
remove link to Reddit from the footer

khalykbek:
GrapheneOS has moved away from Reddit to the combination of self-hosted discussion forum and chat rooms that are bridged across Discord, Telegram and Matrix.


matchboxbananasynergy:
Unsure if we want to do this. Reddit still serves a purpose as it is another place where announcements are posted, so people with a Reddit account can follow it there, along with occasional posts being approved.

thestinger:
That's true, might undo it for now then.

PR #895, opened 2024-03-14
[Usage page] Fix typos

d4rklynk:
I also changed the id for **Navigation mode**


thestinger:
@d4rklynk If you're renaming a section, it needs to be added to redirect.js.

static/usage.html
```diff
@@ -350,7 +350,7 @@ <h2><a href="#contact-scopes">Contact Scopes</a></h2>
                 <h2><a href="#accessibility">Accessibility</a></h2>
 
                 <p>GrapheneOS includes all of the accessibility features from the Android Open
-                Source Project and strives to fill in the gaps from not including Google apps and
+                Source Project and strives to fill in the gaps by not including Google apps and
```
other8026: Based on my understanding, the original wording is correct. There are gaps from not including proprietary Google apps. Gaps aren't being filled by excluding Google apps.

static/usage.html
```diff
@@ -432,7 +432,7 @@ <h3><a href="#updates-settings">Settings</a></h3>
                     features.</p>
 
                     <p>The "Permitted networks" setting controls which networks will be used to perform
-                    updates. It defaults to using any network connection. It can be set to "Non-roaming"
+                    updates. It defaults to use any network connection. It can be set to "Non-roaming"
```
other8026: Grammatically, I think the original is correct? Or it could be changed to "Its default is to use..."

static/usage.html
```diff
@@ -460,11 +460,11 @@ <h3><a href="#updates-security">Security</a></h3>
 
                     <p>The update server isn't a trusted party since updates are signed and verified along
                     with downgrade attacks being prevented. The update protocol doesn't send identifiable
-                    information to the update server and works well over a VPN / Tor. GrapheneOS isn't
+                    information to the update server and works well over a VPN/Tor. GrapheneOS isn't
                     able to comply with a government order to build, sign and ship a malicious update to a
                     specific user's device based on information like the IMEI, serial number, etc. The
                     update server only ends up knowing the IP address used to connect to it and the
-                    version being upgraded from based on the requested incremental.</p>
+                    version being upgraded based on the requested incremental.</p>
```
other8026: I personally think keeping the "from" makes it clearer.

static/usage.html
```diff
@@ -476,9 +476,9 @@ <h3><a href="#updates-disabling">Disabling</a></h3>
 
                     <p>It's highly recommended to leave automatic updates enabled and to configure the
                     permitted networks if the bandwidth usage is a problem on your mobile data connection.
-                    However, it's possible to turn off the update client by going to Settings ➔ Apps,
-                    enabling Show system via the menu, selecting System Updater and disabling the
-                    app.  If you do this, you'll need to remember to enable it again to start receiving
+                    However, it's possible to turn off the update client by going to Settings ➔ Apps ➔ See all xxx apps,
```
other8026: perhaps it's better to use something like ### instead of xxx

static/usage.html
```diff
@@ -618,16 +618,16 @@ <h2><a href="#web-browsing">Web browsing</a></h2>
                 a WebView implementation (GeckoView is not a WebView implementation), so it has to be
                 used alongside the Chromium-based WebView rather than instead of Chromium, which means
                 having the remote attack surface of two separate browser engines instead of only one.
-                Firefox / Gecko also bypass or cripple a fair bit of the upstream and GrapheneOS
+                Firefox/Gecko also bypasses or cripples a fair bit of the upstream and GrapheneOS
                 hardening work for apps. Worst of all, Firefox does not have internal sandboxing
-                on Android. This is despite the fact that Chromium semantic sandbox layer on
+                on Android. Even though Chromium semantic sandbox layer on
```
other8026: Unsure, but if this change is made wouldn't it be better to combine the two sentences? This way, separating the two sentences seems awkward to me.

For example: "Worst of all, Firefox does not have internal sandboxing on Android even though [the] Chromium semantic sandbox layer on Android implemented via the OS..."

static/usage.html
```diff
@@ -1087,18 +1087,18 @@ <h3><a href="#sandboxed-google-play-configuration">Configuration</a></h3>
                     Google Play services along with the Nearby Devices permission for it to have
                     all the access it needs. You need to use the "Google Location Accuracy" link
                     from the sandboxed Google Play configuration menu to access the Google Play
-                    services menu for opting into their network location service, otherwise this
+                    services menu for opting into their network location service, otherwise, this
```
other8026: Is this comma necessary?

static/usage.html
```diff
@@ -1188,13 +1188,13 @@ <h2><a href="#android-auto">Android Auto</a></h2>
                 permission toggles, links to related configuration screens, configuration tips, and
                 links to optional Android Auto dependencies.</p>
 
-                <p>The permission toggles ask for a confirmation before turning on. The
+                <p>The permission toggles ask for a confirmation before turning it on. The
```
other8026: Actually, I think the original should have been "before enabling the permission."

Adding "it" makes it seem like all the toggles are doing are turning themselves on, but they're really enabling/granting permissions.

static/usage.html
```diff
@@ -1208,7 +1208,7 @@ <h2><a href="#android-auto">Android Auto</a></h2>
                 allowed notification access. The notification access settings are linked below the
                 permission toggles.</p>
 
-                <p>In order to show up in the Android Auto car interface, apps have to be installed
+                <p>For Android Auto to show up in the car interface, apps have to be installed
```
other8026: I think the original is more accurate. This section is about apps showing up, not Android Auto.

static/usage.html
d4rklynk: "Its default is to use..." (as you wrote) seems better imo

static/usage.html
d4rklynk: I would say the opposite, but tbf I trust you since english is not my main language.

static/usage.html
d4rklynk: I guess you mean replacing "from" from "based" right ?

static/usage.html
d4rklynk: Maybe, I was wondering the same 🤔

static/usage.html
d4rklynk: Yeah, it is better indeed.

static/usage.html
d4rklynk: Maybe not indeed.

static/usage.html
d4rklynk: Oh ok, I just wanted to make sentences less wordy tbf. In that case yeah, you can keep the original.

static/usage.html
d4rklynk: It seems redundant to me to have both words, I thought it was a mistake actually.

d4rklynk:
@other8026 Just to be clear, you want **me** to make the changes right ? Sry if it seems obvious.

static/usage.html
d4rklynk: The sentence should be written again yeah. "before enabling the permission" seems clearer to me.

static/usage.html
other8026: I mean I don't think this part needs to be changed.

static/usage.html
other8026: Yes, I think keep the original unchanged.

static/usage.html
d4rklynk: I really find it weird, so weird that I don't really understand the end of the sentence with "from" in it. That could be me tbf. 

If you think so, let's keep it then.

other8026:
@d4rklynk I think that's best so anyone else who wants to review the changes can more easily do so

static/usage.html
```diff
@@ -1247,7 +1247,7 @@ <h2><a href="#banking-apps">Banking apps</a></h2>
 
                 <p>The hardware attestation feature is part of the Android Open Source Project and
                 is fully supported by GrapheneOS. SafetyNet attestation chooses to use it to
-                enforce using Google certified operating systems. However, app developers can use
+                enforce using Google-certified operating systems. However, app developers can use
```
other8026, line 1273: In the past, the project has always written this as "Google certified", so I think it's best to keep it that way here too.

other8026 (commented):
Looking over everything again, I found just this one that I missed. Sorry for the trouble!

static/usage.html
d4rklynk: No problem!

d4rklynk:
Let me know if you find any other issues !

other8026 (approved):

static/usage.html
d4rklynk: @other8026  Sry to bother you again, but what about making this sentence more like so? And I'll stop here if you don't agree.

"The update server ultimately only gains knowledge of the IP address used for connection and the version being upgraded from, based on the requested incremental update."

other8026: @d4rklynk In short, I think this section should remain unchanged for now.

It's very important information and it needs to be 100% accurate as well as very clear. I think the feedback that you've given is helpful. It's very likely that others who have read this section have also found the wording to be unclear, which means it should be addressed.

Due to its importance, I think any change to this section should be its own pull request and the wording should be reviewed by multiple people.

d4rklynk: I understand, thanks for explaining to me.

thestinger:
@d4rklynk It conflicts with other changes that were made now. It will need to be rebased.

other8026 (approved):

thestinger:
This really needs to be rebased in order to merge it.

PR #931, opened 2024-05-15
Revert "Fix branch for akita kernel to 14-akita"

joker22008:
This reverts commit 3332fa659663f51efe4653bbf05278f9b501c284.


thestinger:
It's meant to be `14-akita` for now, until the device branch is gone. It uses different revisions of repositories than shusky despite sharing same of the same repositories.

Issue #860, opened 2024-02-02
Document NodePing status page


orazioedoardo:
https://nodeping.com/reports/statusevents/3FL97F6RRQHJIOWL exists but it’s not documented anywhere on the site. Could be added in the footer or on https://grapheneos.org/articles/grapheneos-servers.

PR #947, opened 2024-06-06
Add footer link to NodePing status page

mackalex:
### Description
As the PR title reads. Adds the NodePing status page as a link in the footer.

### Related Issue
#860


thestinger:
I think this belongs somewhere else like the GrapheneOS servers page.

PR #963, opened 2024-06-27
Static page tweaks

de0u:
This makes a few content-neutral changes to static/faq.html, including adding a period at the end of a sentence and adding some commas.


de0u:
Source for the comma after "etc.": https://jakubmarian.com/comma-before-and-after-etc/

PR #974, opened 2024-07-14
use Protobuf library for Python from PyPI

ghost:


static/build.html
```diff
@@ -377,14 +378,20 @@ <h3><a href="#extracting-vendor-files-for-pixel-devices">Extracting vendor files
 
                     <p>The below commands need to only be run once to initially create a working environment.</p>
 
-                    <pre>yarn install --cwd vendor/adevtool/
+                    <pre>python3 -m venv .venv
```
thestinger: Just call it venv instead of .venv.

thestinger (changes requested):
It should be optional for Linux distributions with a consistently up-to-date package.

PR #982, opened 2024-07-20
Add build instructions for simple prebuilt apps

TanguyP:
As suggested in [this](https://github.com/GrapheneOS/PdfViewer/issues/373) PDF Viewer issue, it could be useful to document on the website the build process for apps such as PDF Viewer. Though the process is quite simple, having it documented lowers the barrier for contributions.

Therefore, this PR proposes a first version of such documentation, which I've tried to keep high-level and mostly OS-agnostic. Once we have this, we can modify the READMEs of repositories which comply with this build process (PDF Viewer, Apps, etc.) to point to this documentation.

### DIFF
--- a/static/build.html
+++ b/static/build.html
@@ -1035,7 +1035,7 @@ git checkout <var>VERSION</var></pre>
                     standard <code>gradle</code> Android plugin build system. You can also build those apps from the source code by following these steps:
                     </p>
                     <ul>
-                        <li>Install Gradle, OpenJDK 17 for developers (which should be called <code>java-17-openjdk-devel</code> on Linux), Android Studio, and npm.</li>
+                        <li>Install Gradle, OpenJDK 21 for developers (which should be called <code>java-21-openjdk-devel</code> on Linux), Android Studio, and npm.</li>
                         <li>In the project you want to build, create a <code>local.properties</code> file and inside it, set the path to your JDK e.g. <code>sdk.dir=/home/jane.doe/Android/Sdk</code></li>
                         <li>In Android Studio's Device Manager, make sure you have a device running (virtual device, or physical device paired using wi-fi)</li>
                         <li>From a console in your project folder, run <code>gradle installDebug</code>. This will install the app from the source code, alongside any prebuilt version of the app which you may already have.</li>
--- a/static/build.html
+++ b/static/build.html
@@ -1032,7 +1032,14 @@ git checkout <var>VERSION</var></pre>
                     <p>The official releases of our App Store, Auditor, Camera and PdfViewer apps are
                     bundled as apks into external/ repositories. The no-code AppCompatConfig and
                     GmsCompatConfig apps are done the same way. These are built and signed with the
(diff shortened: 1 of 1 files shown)


thestinger:
@TanguyP Can you rebase this and change it to OpenJDK 17 instead of 21? We actually use 21 for Gradle and 17 for compilation.

We rebased the history to deter people responding to older wording than what we published on the site for an article. I know it's annoying.

TanguyP:
> @TanguyP Can you rebase this and change it to OpenJDK 17 instead of 21?

Presumably you meant the other way around :) it's done. Do let me know if there are any other concerns with the PR.

thestinger:
Yes, I meant the other way around.

static/build.html
```diff
@@ -1032,7 +1032,14 @@ <h3><a href="#prebuilt-apps">Prebuilt apps</a></h3>
                     <p>The official releases of our App Store, Auditor, Camera and PdfViewer apps are
                     bundled as apks into external/ repositories. The no-code AppCompatConfig and
                     GmsCompatConfig apps are done the same way. These are built and signed with the
-                    standard <code>gradle</code> Android plugin build system.</p>
+                    standard <code>gradle</code> Android plugin build system. You can also build those apps from the source code by following these steps:
+                    </p>
+                    <ul>
+                        <li>Install Gradle, OpenJDK 21 for developers (which should be called <code>java-21-openjdk-devel</code> on Linux), Android Studio, and npm.</li>
```
thestinger, line 1038: This is only the name of the package on Debian distributions and isn't the name on Arch Linux or elsewhere. On Arch Linux, the package name is `jdk21-openjdk` by comparison with `jre21-openjdk` as the subset without all the development tools. Debian has a standard devel split they use but it's not clear exactly what they did in this case. Arch Linux normally doesn't split headers, etc. into a separate package but the JDK is a special case.

static/build.html
```diff
@@ -1032,7 +1032,14 @@ <h3><a href="#prebuilt-apps">Prebuilt apps</a></h3>
...
+                        <li>In the project you want to build, create a <code>local.properties</code> file and inside it, set the path to your JDK e.g. <code>sdk.dir=/home/jane.doe/Android/Sdk</code></li>
```
thestinger, line 1039: It's a lot easier to export `ANDROID_HOME` in your environment so that Gradle creates `local.properties` itself for each project.

thestinger (changes requested):

static/build.html
TanguyP: Good point, thanks. Would this be ok?
```suggestion
                        <li>Install Gradle, Android Studio, npm, and OpenJDK 21 (<code>java-21-openjdk-devel</code> package on Debian distributions, <code>jdk21-openjdk</code> on Arch Linux distributions...).</li>
```

static/build.html
TanguyP:
```suggestion
                        <li>Export the <code>ANDROID_HOME</code> environment variable and make it point to your JDK folder</li>
```

PR #983, opened 2024-07-22
Explain how to fix WebUSB of Chromium on Ubuntu

Kixunil:
The documentation previously claimed that snap-packaged chromium has WebUSB broken on Ubuntu. This is not entirely accurate as it is just disabled. To enable it one needs to run one simple command after which flashing from Web starts working. (I've tested this.)

This commit updates the website to change the claim into explaining which command is needed.

### DIFF
--- a/static/install/web.html
+++ b/static/install/web.html
@@ -130,7 +130,7 @@
                 <p>Officially supported browsers for the web install method:</p>
 
                 <ul>
-                    <li>Chromium (outside Ubuntu, since they ship a broken Snap package without working WebUSB)</li>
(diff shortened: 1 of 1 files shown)


thestinger:
@Kixunil This has been undone for now because it doesn't fit the way we want the install guide to work. You should really get Ubuntu to fix this problem instead of having an extra command included in a place that's not meant to have commands.

Kixunil:
I'm not sure if anyone can get them to "fix" it if they left it intentionally disabled for supposedly better security. I thought that at least providing a solution would be better than leaving people hanging since there aren't better alternatives. Would you be open to a different PR that places the command in a different section? E.g. before/after `fwupd` workaround.

thestinger:
@Kixunil Yes, it should probably be in a section like that workaround.

PR #984, opened 2024-07-23
Explain how to enable USB on snap-packaged Chromium

Kixunil:
The installer documentation previously claimed that USB on Chromium packaged by Ubuntu is broken. This was not accurate as it's just disabled and can be easily enabled.

This commit removes/corrects the statements about Snap being problematic and adds a section that explains which command can fix the problem.

### DIFF
--- a/static/install/web.html
+++ b/static/install/web.html
@@ -130,14 +130,15 @@
                 <p>Officially supported browsers for the web install method:</p>
 
                 <ul>
-                    <li>Chromium (outside Ubuntu, since they ship a broken Snap package without working WebUSB)</li>
+                    <li>Chromium</li>
                     <li>Vanadium (GrapheneOS)</li>
                     <li>Google Chrome</li>
                     <li>Microsoft Edge</li>
                     <li>Brave (with Brave Shields disabled, since it caps storage usage at a low value to avoid fingerprinting available storage)</li>
                 </ul>
 
-                <p>You should avoid Flatpak and Snap versions of browsers, as they're known to cause issues during the installation process.</p>
+                <p>You should avoid Flatpak and Snap versions of browsers, as they're known to cause issues during the installation process.
+                However some of them can be solved.</p>
 
                 <p>Make sure your browser is up-to-date before proceeding.</p>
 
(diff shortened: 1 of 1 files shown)


static/install/web.html
```diff
@@ -194,6 +195,17 @@ <h2><a href="#flashing-as-non-root">Flashing as non-root</a></h2>
...
+            
+            <section id="enabling-webusb-on-snap-packaged-chromium">
+                <h2><a href ="#enabling-webusb-on-snap-packaged-chromium">Enabling WebUSB on snap-packaged Chromium</a></h2>
+
+                <p>Ubuntu and some derived distributions ship Chromium packaged in Snap. This has access to USB disabled by default.
+                You can enable it using this command:</p>
+
+                <pre>sudo snap connect chromium:raw-usb</pre>
+
+                <p>You might need to restart the browser afterwards.</p>
```
thestinger: It would be good to confirm if this is required or not to either say it's required or omit it.

Kixunil: Can I somehow check the USB connection from web without having to try unlocking or whatever? The phone I used is already installed and I don't want to mess with it too much.

thestinger: Use the lock button to see if it connects and gets an error from it already being locked.

Kixunil:
```suggestion
```

Kixunil: Found out restart is not really needed.

ghost:
Can confirm that `sudo snap connect chromium:raw-usb` makes it possible to successfully flash GrapheneOS using the Ubuntu Chromium snap 126.0.6478.182 on Ubuntu 24.04.

thestinger:
This won't be merged. The web install guide should not have any CLI commands. The existing desktop Linux fwupd command will be removed, and this won't be added. Ubuntu should fix their Chromium package to have working WebUSB without requiring a CLI command to work around it. If they want to have a permission system for USB, great, they should implement one which requests access to a USB device and the user chooses it. The user needing to open a terminal and run a command to grant full USB access to all USB devices is neither usable or secure. We'll drop Ubuntu as a supported OS sooner than we do that.

Kixunil:
I don't disagree but it's kinda sad that users won't see the potential solution. Granting access to all USB devices is only a problem if chromium itself gets compromised (websites already ask permissions) which should be pretty unlikely. I agree it's a UX nightmare, though almost every Linux user knows how to use terminal. Well, the technical ones will hopefully find this issue.

thestinger:
It's not very reasonable that we have to document a bunch of CLI workarounds for desktop Linux. They should not be making Chromium broken by default with a CLI workaround required. There should be a user interface for it which prompts for it...

Kixunil:
Side note: Ubuntu-based distributions such as Pop!OS also inherit this problem.

I agree, hopefully I'll find some time to file an issue at Ubuntu.

thestinger:
We can potentially include info about it but I don't want to add a new section and this wasn't doing that correctly.

PR #986, opened 2024-07-28
web-install + cli: updated guide docs on removing non-stock key while in fastboot

jmvermeulen:
The current web installer documentation is not clear about the required phone state while removing the non-stock key.
The phone should be in *fastboot* mode and bootloader should be unlocked. Otherwise it's not detected by WebUSB.

### DIFF
--- a/static/install/cli.html
+++ b/static/install/cli.html
@@ -650,8 +650,8 @@ curl -O https://releases.grapheneos.org/<var>DEVICE_NAME</var>-factory-<var>VERS
 
                     <p>The GrapheneOS factory images flash a non-stock Android Verified Boot key which
                     needs to be erased to fully revert back to a stock device state. Before flashing the
-                    stock factory images and before locking the bootloader, you should erase the custom
-                    Android Verified Boot key to untrust it:</p>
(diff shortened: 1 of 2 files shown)


thestinger:
Needs to be changed in both guides, not just web.

dev-goalz:
> Needs to be changed in both guides, not just web.

Updated the cli guide too and updated the PR.

static/install/cli.html
```diff
@@ -650,8 +650,8 @@ <h3><a href="#replacing-grapheneos-with-the-stock-os">Replacing GrapheneOS with
 
                     <p>The GrapheneOS factory images flash a non-stock Android Verified Boot key which
                     needs to be erased to fully revert back to a stock device state. Before flashing the
-                    stock factory images and before locking the bootloader, you should erase the custom
```
thestinger, line 653: Instead of "bring", it would make more sense to say "boot".

thestinger (changes requested):

static/install/cli.html
dev-goalz: Done

PR #1009, opened 2024-09-01
Remove mention that fwupd workaround is not needed for Pixel 4a and later from webinstall docs

rocodes:
Hi, thanks for all of your work on GrapheneOS and the excellent documentation.

I am proposing to remove the line in the webinstall docs that indicates that the fwupd.service workaround is no longer required for 4a and later devices; I recently went through the webinstall instructions with a newer Pixel device on Debian bookworm, which still has fwupd < 1.9.10, and required the workaround to proceed.

### DIFF
--- a/static/install/web.html
+++ b/static/install/web.html
@@ -198,8 +198,6 @@
             <section id="working-around-fwupd-bug-on-linux-distributions">
                 <h2><a href="#working-around-fwupd-bug-on-linux-distributions">Working around fwupd bug on Linux distributions</a></h2>
 
-                <p><em>This is no longer required for the Pixel 4a (5G) and later.</em></p>
-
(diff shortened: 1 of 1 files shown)


thestinger:
Which device did you use? It says it's no longer required for the Pixel 4a (5G) and later which is a later device than the Pixel 4a.

rocodes:
I used a Pixel 8.

rocodes:
If this workaround is not expected to be needed on (Pixel 4a...current), I am happy to open a ticket

jermanuts:
Can confirm that you don't need to disable `fwupd`

rocodes:
This is a pretty old ticket, the fwupd version and device versions I referenced are now outdated, I'm just going to close it.

PR #1010, opened 2024-09-02
update Android Studio URL

ghost:


thestinger:
Can you explain why this is better?

ttbb5566 (approved):

thestinger:
See https://github.com/golang/go/issues/22648 as an counter-example. I think I'd prefer not to switch to the redirector URL.

PR #1024, opened 2024-09-23
restoring a previous paragraph

b068931cc450442b63f5b3d276ea4297:
In commit https://github.com/GrapheneOS/grapheneos.org/commit/c7e634364aa2e2774a9cdbe5172f0983f4d827ba#diff-81aefcb9b7a401823bfb4558719f8aa4dbce2ea47cad509c7cc2111abf31d7d5L108 the paragraph for the requirements was replaced by those for the web installer.

### DIFF
--- a/static/install/cli.html
+++ b/static/install/cli.html
@@ -105,9 +105,7 @@
             <section id="prerequisites">
                 <h2><a href="#prerequisites">Prerequisites</a></h2>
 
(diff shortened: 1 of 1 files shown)


thestinger:
It was intended to add the web installer mention here to explain that there's a way to use a broader range of devices but the initial text wasn't changed for the CLI install. Fixed this now.

PR #1028, opened 2024-09-30
add bootloader screen sample

ghost:


matchboxbananasynergy:
I really like how this looks, but if we're planning to add more "visual aid" to the install pages, we'll eventually need to use images, right? (i.e. to show the OEM unlocking toggle)

If we're planning to use images eventually, perhaps it would make sense to use images from the get-go?

thestinger:
The issue with an image is that we can't take a screenshot of it so it's not easily possible to obtain a proper image to use. I wouldn't want to use a photo taken with a camera. This approach does seem a bit strange and I'm not sure I want to do it this way. It mimics how it looks on the device but it doesn't actually look the same.

thestinger:
I don't think supporting text-only browsers really matters.

matchboxbananasynergy:
I also agree with supporting text-only browsers not mattering, I doubt that any significant amount of people is using them.

That said, I do really like how this looks, and given that this is something we can't take a screenshot of, I think this might be the best way to do it.

People do seem to struggle with finding this screen or thinking they can proceed when they're not at it, so a visual representation of what they should be seeing on their phone/tablet before proceeding will be very helpful.

I think it makes sense to merge this.

de0u:
I like this, but we have had some instances where users see the word "Start" next to the power button and they press the power button in order to get the installation process started, though of course activating "Start" means that the installation process cannot start.

So I would argue for enlarging the fake image to show the button menu and somehow (text and/or diagram) make it clear that pushing the power button to engage the "Start" function must **_not_** be done.

thestinger:
Unfortunately, I think my lukewarm response to this led to them deleting their account and 4 open pull requests. I do agree we lack a way to use a proper image for this due to inability to take a screenshot but that is what I would prefer if it was possible. I would like to improve the install guides with a visual reference for this but I'm not sure this is the right approach. I would prefer an actual image, and a far from pixel perfect one taken with a camera would be a suitable starting point. Ideally we could have a screenshot. It would be a waste of time to try to render exactly the same thing they do into an image especially since it will change over time.

PR #1032, opened 2024-10-06
Add section on devices with GrapheneOS preinstalled

matchboxbananasynergy:
This PR adds a section regarding devices with GrapheneOS preinstalled. It is meant to do the following:

1. Make it clear that we're not affiliated with any companies selling devices with GrapheneOS preinstalled.
2. Reiterate that our recommendation is for people to do it themselves with our web installer.
3. Help users who decide to purchase a device from a third-party anyway to make an informed decision and ensure they're running unmodified and clean GrapheneOS.


thestinger:
Merged with conflict fixed.

PR #1049, opened 2024-11-02
Update usage eSIM section

WaleedMortaja:


thestinger:
This could be reworded but I want to do it differently from this.

thestinger:
https://github.com/GrapheneOS/grapheneos.org/commit/92b9ebfa74529c5721820387475bbc30e4b6cf34

WaleedMortaja:
@thestinger What about [this line](https://github.com/GrapheneOS/grapheneos.org/commit/92b9ebfa74529c5721820387475bbc30e4b6cf34#diff-39b0493c6a89dcdf21af5249d7e3fa43b87b4acce652d2e76e54c30849b5575aL1167): ``` However, in order to manage and add eSIMs, proprietary Google functionality is needed. ``` ?
proprietary Google functionality is not really needed anymore.

thestinger:
It is still proprietary functionality written by Google, we just avoid depending on Google Play or Google services. We could change that wording to avoid misleading people.

WaleedMortaja:
@thestinger Before the 2024012600 update, It was required to enable Sandboxed Google Play in order to mange the eSIMs, but not any more.

I would suggest something like:
```
By default, GrapheneOS has always shipped with baseline support for eSIM, where users can use any eSIMs installed previously on the device. 

Managing and adding eSIMs is a (  / contains some?) proprietary Google functionality. This is fully disabled by default.

eSIM support on GrapheneOS doesn't require any dependency on Google Play, so it is not required to have the Sandoxed Google Play installed.

GrapheneOS never shares data to Google Play even when installed. It won't connect to a Google service unless the carrier uses one themselves.
```

What do you think?

PR #1051, opened 2024-11-06
add formatter

cookiell546:


thestinger:
@cookiell546 Please leave the `.github` files as is for now because we would want to change them across all repositories ourselves if we were going to switch them to 4 space tabs. Also, please use `100` as the column width rather than the default `80` which we consider too small for modern development environments. We use `100` but we make exceptions where going past it is more readable.

cookiell546:
Done.

PR #1066, opened 2024-11-30
Mikegilchrist patch 1 add note about not using FF and reformat other browser related comments

mikegilchrist:
Suggest modifying instructions to make incompatibility of FF with web install.

### DIFF
--- a/static/install/web.html
+++ b/static/install/web.html
@@ -139,13 +139,15 @@
                     <li>Brave (with Brave Shields disabled, since it caps storage usage at a low value to avoid fingerprinting available storage)</li>
                 </ul>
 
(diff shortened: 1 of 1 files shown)


thestinger:
It would be better to just detect that it's a browser without WebUSB and display an error. I'd rather not add more text for this.

mikegilchrist:
Understand your perspective, but I suspect you could save a consistent stream of users from frustration by implementing this imperfect, but currently available, solution.

thestinger:
The page already says WebUSB is required and has a list of supported browsers. The only improvement that's acceptable is showing an error message when WebUSB is unavailable. Repeating the same things multiple times in the text only makes it longer which means fewer people reading it.

PR #1069, opened 2024-12-03
Update process-templates

ZeroDarkThunder:
Reorganized Python to a more pythonic structure making it easier to read and maintain in the future

### DIFF
--- a/process-templates
+++ b/process-templates
@@ -1,16 +1,17 @@
 #!/usr/bin/env python3
 
-from jinja2 import FileSystemLoader, Environment
-from pathlib import Path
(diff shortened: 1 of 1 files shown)


process-templates
```diff
@@ -1,17 +1,18 @@
 #!/usr/bin/env python3
 
-from jinja2 import FileSystemLoader, Environment
-from pathlib import Path
+ROOT_DIR = Path(sys.argv[1])
```
thestinger, line 3: Doesn't look right. It depends on the imports below and this isn't the normal style.

ZeroDarkThunder: Okay

PR #1070, opened 2024-12-03
Update generate-feed

ZeroDarkThunder:
slight reorg of imports


thestinger:
The current order seems to follow PEP8 already.

PR #1071, opened 2024-12-04
Fixed "Changes since" release codes

An-anonymous-coder:
2024120200 and 2024112700 both said "Changes since the 2024111700 release:", which is incorrect.


An-anonymous-coder:
2022080300 and 2019.11.04.23 also have this issue. Up to you if you want to fix those.

thestinger:
@An-anonymous-coder Can you submit both of those too?

PR #1080, opened 2024-12-16
fix broken same-page link for VPN leak blocking

jermanuts:
https://github.com/GrapheneOS/grapheneos.org/pull/1034 missed it


jermanuts:
sorry wrong branch, will open new one

PR #1084, opened 2024-12-17
Clarify required pwd for path export commands

cleeyv:
I just helped someone who was blocked on the CLI install because they didn't realize they needed to be in a specific directory (the extracted platform-tools zip) for the path export commands for the standalone platform-tools to work properly. This addition to the documentation could help prevent others from getting blocked with the same problem. 

This person (and most people) would have had an easier time with the web installer, but this didn't work for them because they were using ubuntu and didn't realize that chromium was installed with a snap with web-usb disabled. PR #984  being merged would have allowed them to use the web installer which probably would have been better overall.

### DIFF
--- a/static/install/cli.html
+++ b/static/install/cli.html
@@ -258,7 +258,9 @@ tar xvf platform-tools_r35.0.2-darwin.zip</pre>
 tar xvf platform-tools_r35.0.2-win.zip</pre>
 
                     <p>Next, add the tools to your <code>PATH</code> in the current shell so they can be
-                    used without referencing them by file path, enabling usage by the flashing script.</p>
+                    used without referencing them by file path, enabling usage by the flashing script.
+		    The following commands must be run from within the extracted <code>platform-tools_r35.0.2-*.zip</code>
+		    directory.</p>
 
                     <p>On Debian, Ubuntu and macOS:</p>
 


thestinger:
This is incorrect and will break the installation guide because the export command uses a path to the extracted directory:

```
export PATH="$PWD/platform-tools:$PATH"
```

The exact set of commands in the CLI install guide already work perfectly. Deviating from the guide is the issue. If people follow this new instruction you've added, it won't work for them unless they also change the other instructions. It also really doesn't make any sense to download and extract the factory images within the platform-tools directory as this will end up having people do if they follow it and know how to change the directory.

> were using ubuntu and didn't realize that chromium was installed with a snap with web-usb disabled

Our web installer guide already explicitly says not to use Ubuntu's Chromium package.

> PR https://github.com/GrapheneOS/grapheneos.org/pull/984 being merged would have allowed them to use the web installer which probably would have been better overall.

This won't be merged. The web install guide should not have any CLI commands. The existing desktop Linux fwupd command will be removed, and this won't be added. Ubuntu should fix their Chromium package to have working WebUSB without requiring a CLI command to work around it. If they want to have a permission system for USB, great, they should implement one which requests access to a USB device and the user chooses it. The user needing to open a terminal and run a command to grant full USB access to all USB devices is neither usable or secure. We'll drop Ubuntu as a supported OS sooner than we do that.

cleeyv:
You are correct. I had not followed the instructions very carefully and had extracted the zip file with a tool other than bsdtar, resulting in a different directory structure which could have benefited from the proposed clarification. When bsdtar is used the current instructions work without any clarification.

I saw PR #984 when I was preparing this PR and realized it was the same problem my friend had encountered but didn't go back and read through the web installer guide before mentioning it here.

Sorry to bother you with an unhelpful PR! Keep up the good work.

PR #1085, opened 2024-12-20
Removed US McDonalds app from list of apps banning GrapheneOS

diabler-ie:
Removed US McDonalds app from list of apps banning GrapheneOS as it appears they have made changes that allow the app to work for us :)


thestinger:
Thanks, already made the change before seeing the email for this.

PR #1086, opened 2024-12-21
list "symptoms" of successful entry to fastboot mode

de0u:
The discussion forum has seen quite a few instances of users who think they have entered fastboot but haven't, or who enter fastboot but accidentally leave it by invoking the "Start" action (because they wish to start the installation process and "Start" looks attractive).  Here are two examples of users indicating that explicit directions would be helpful:
* https://discuss.grapheneos.org/d/10515-installation-error-unlocking-bootloader/29
* https://discuss.grapheneos.org/d/10515-installation-error-unlocking-bootloader/26

This commit spells out how a user can be sure that fastboot is active and also that it is important to **_not_** invoke the "Start" action.

### DIFF
--- a/static/install/cli.html
+++ b/static/install/cli.html
@@ -333,6 +333,11 @@ Installed as /home/username/platform-tools/fastboot</pre>
                 <p>Alternatively, turn off the device, then boot it up while holding the volume
                 down button during the boot process. You can either boot it with the power button
                 or by plugging it in as required in the next section.</p>
+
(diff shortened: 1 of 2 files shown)


thestinger:
This part of the install guide is before users unlock so it won't say that it's unlocked yet unless it was already unlocked before following our guide.

de0u:
Oops!  My apologies.  Fixed, I think.

PR #1090, opened 2024-12-27
add HEK Service-App to attestation compatibility guide

reifenrath-dev:
Recently I learned that the HEK service app from a German health insurance company also seems to use Play Integrity API and therefore cannot be used on GrapheneOS, so I would like to add it to the list.

### DIFF
--- a/static/articles/attestation-compatibility-guide.html
+++ b/static/articles/attestation-compatibility-guide.html
@@ -151,6 +151,7 @@
                     <li><a href="https://play.google.com/store/apps/details?id=com.revolut.business">Revolut Business</a></li>
                     <li><a href="https://play.google.com/store/apps/details?id=com.mcdonalds.mobileapp">McDonald's</a> (International app used for many but not all countries not including the US)</li>
                     <li><a href="https://play.google.com/store/apps/details?id=de.tk.tkapp">TK-App</a></li>
+                    <li><a href="https://play.google.com/store/apps/details?id=de.hek.serviceapp">HEK Service-App</a></li>
                 </ul>
 
                 <p>In addition to leaving feedback for these apps on the Play Store, file support


thestinger:
We need to confirm this somehow. Do you have a screenshot? Can you get logs from View logs in the Settings app? Might need from the OS in addition to the app but the app logs might be enough.

reifenrath-dev:
@thestinger thanks for your response. I tried for some hours now to find some relevant parts in the app and system log. Unfortunately I'm not experienced in android development. I also tried to look at the logs of the Revolut App to get an idea of what to look for. I could not find anything saying "Play Integrity API" or similar (which I could also not find for the Revolut App, tough). However, the following lines could be relevant:
```
12-27 20:53:01.039  6607  6607 W libc    : Access denied finding property "ro.product.name_for_attestation"
12-27 20:53:01.039  6607  6607 W libc    : Access denied finding property "ro.product.device_for_attestation"
12-27 20:53:01.039  6607  6607 W libc    : Access denied finding property "ro.product.manufacturer_for_attestation"
...
12-27 20:53:01.038  6607  6607 W main    : type=1400 audit(0.0:16038): avc:  denied  { read } for  name="u:object_r:build_attestation_prop:s0" dev="tmpfs" ino=11280 scontext=u:r:untrusted_app:s0:c124,c256,c522,c768 tcontext=u:object_r:build_attestation_prop:s0 tclass=file permissive=0
12-27 20:53:01.039  6607  6607 W libc    : Access denied finding property "ro.product.brand_for_attestation"
12-27 20:53:01.039  6607  6607 W libc    : Access denied finding property "ro.product.model_for_attestation"
...
12-27 20:53:03.297  6607  6669 E GoogleApiManager: Failed to get service from broker. 
12-27 20:53:03.297  6607  6669 E GoogleApiManager: java.lang.SecurityException: Unknown calling package name 'com.google.android.gms'.
...
12-27 20:53:03.304  6607  6668 W CondFlagRegistrar: Failed to register com.google.android.gms.providerinstaller#com.google.android.gms
12-27 20:53:03.304  6607  6668 W CondFlagRegistrar: ehvp: 17: 17: API: Phenotype.API is not available on this device. Connection failed with: ConnectionResult{statusCode=DEVELOPER_ERROR, resolution=null, message=null}
```

Could you point me to what I have to look out for or if you have the time look at the full log?

The full HEK app log:
[HEK log 53ae20b42e5e.txt](https://github.com/user-attachments/files/18263993/HEK.log.53ae20b42e5e.txt)

thestinger:
It seems like they're likely doing this another way than Play Integrity so it doesn't really fit into the existing section, but could be added in another list there. It would be good to know why it's not working though so we can properly complain. Perhaps they're using key attestation but only allowing stock OS.

reifenrath-dev:
It seems like they fixed this (intentionally or not). At least I can start the app. I cannot verify however, if the login works and everything after the login since I am not a customer.

Issue #1088, opened 2024-12-26
Web Installer guide has sightly outdated information regarding fwupd on Debian/Ubuntu


rslurry:
Currently the Web Installer guide says that Debian and Ubuntu have an outdated fwupd package that causes issues with the install, and that problem is solved by fwupd versions >=1.9.10.  I just checked Ubuntu 24.04.1 and it has fwupd version 1.9.24 by default.  I can't confirm if this is also the case for the latest Debian release, but I suspect it might not be given it lags 1-2 years behind.

I recommend updating the language of the guide accordingly (e.g., "Debian stable and Ubuntu **may** have an outdated fwupd package ...").

I think it would also be good to include a command to check the installed version, e.g., `apt-cache policy fwupd`, to make it easier for less technical users to check and know whether or not they need to worry about this.

thestinger:
It's going to remain as it is because it appears the newer versions of fwupd may still cause problems.

rslurry:
I used a totally fresh install of Ubuntu 24.04.1 and had zero problems installing GOS.  If fwupd 1.9.24 caused problems, wouldn't that have prevented me from completing the install without stopping the service as instructed in the current Web Installer guide?  The absolute language in the current guide is clearly not 100% correct and should be corrected, perhaps not exactly as I suggested, but at a minimum using more reasonable language, e.g., "may need to stop the service".

thestinger:
https://packages.debian.org/search?keywords=fwupd
https://launchpad.net/ubuntu/+source/fwupd

Ubuntu 24.10 and 24.04 have a new enough version, but the other still supported stable releases don't have it. We could change the wording to be specific about which versions require it but that will continue making the guide more complex. Another option is dropping support for Debian and Ubuntu versions older than 24.04.

rslurry:
The guide mentions that fwupd 1.9.10 and later do not have this issue, is that confirmed?  Taking a step back, I think the guide could be made more general and therefore easier to maintain while addressing this difference between Ubuntu 24.04+ vs. older versions without having to specifically identify them, as there are other Debian- and Ubuntu-based distributions (e.g., Kali, PopOS, SteamOS) that are not mentioned as supported in the current guide but would likely work for the install given their origins.  The section could instead say something like,

[...]

PR #1091, opened 2024-12-30
Generalized install info related to fwupd bug on Debian-based distros

rslurry:
Updated Web Install info related to issue #1088 about the fwupd bug in Debian and some Ubuntu versions.  This avoids unnecessarily stopping fwupd on releases that don't have the bug, and it keeps the guide generalized for applicability to other Debian- and Ubuntu-derived distros.  It won't need to be revisited until years from now when Debian and Ubuntu entirely drop support for the current releases that use the outdated fwupd version.

### DIFF
--- a/static/install/web.html
+++ b/static/install/web.html
@@ -54,7 +54,7 @@
                     <li><a href="#prerequisites">Prerequisites</a></li>
                     <li><a href="#enabling-oem-unlocking">Enabling OEM unlocking</a></li>
                     <li><a href="#flashing-as-non-root">Flashing as non-root</a></li>
-                    <li><a href="#working-around--bug-on-linux-distributions">Working around  bug on Linux distributions</a></li>
+                    <li><a href="#working-around-fwupd-bug-on-linux-distributions">Working around  bug on Linux distributions</a></li>
                     <li><a href="#booting-into-the-bootloader-interface">Booting into the bootloader interface</a></li>
                     <li><a href="#connecting-device">Connecting the device</a></li>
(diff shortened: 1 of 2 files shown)


thestinger (changes requested):
This needs to be kept in sync between the CLI and web install guides.

static/install/web.html
```diff
@@ -54,7 +54,7 @@ <h2><a href="#table-of-contents">Table of contents</a></h2>
                     <li><a href="#prerequisites">Prerequisites</a></li>
                     <li><a href="#enabling-oem-unlocking">Enabling OEM unlocking</a></li>
                     <li><a href="#flashing-as-non-root">Flashing as non-root</a></li>
-                    <li><a href="#working-around-fwupd-bug-on-linux-distributions">Working around fwupd bug on Linux distributions</a></li>
+                    <li><a href="#working-around--bug-on-linux-distributions">Working around  bug on Linux distributions</a></li>
```
thestinger: This looks accidental.

thestinger (changes requested):

static/install/web.html
rslurry: Huh, I didn't intentionally touch anything there - must have been an accident when Ctrl+F'ing for fwupd.  Thanks for catching that.

rslurry:
Alright, that accidental deletion was fixed, and the CLI guide was updated to match.

rslurry:
Hey there, just circling back now that the holidays have passed.  Are there any other changes you'd like to see before merging this PR?

PR #1121, opened 2025-02-14
Update features.html: typo fix

nikelborm:


thestinger:
The existing way this was written was correct. We turn it on by default so that all Vanadium users have it on in practice to avoid people turning it on and differentiating themselves.

nikelborm:
What?

nikelborm:
You want to do the opposite aren't you? You just exactly make people identifiable by differentiating this value from the masses (and the masses have this turned off)

Mozilla removed it completely from Firefox, to make users less identifiable and prevent them from enabling it.

thestinger:
The masses don't use Vanadium. It's not possible for us to pretend Vanadium is something else due to features like having the JIT disabled by default. The goal of anti-fingerprinting is making all Vanadium users look the same.

nikelborm:
I thought the goal was to make all vanadium users to look like the most average browser user ever with the most common markers contributed to overall fingerprint

thestinger:
That's not how fingerprinting works. The individual properties being checked don't exist in isolation.

nikelborm:
Fingerprint is hash sum of all parameters. The higher amount of static parameters, the higher the chance of collisions. There are different fingerprint algorithms that take into account different parameters. Some of them care about JITness, some of them don't.

Having static do-not-track (regardless if it's on/off) is better than having user-changeable do-not-track because you reduce the chance of individuals using vanadium to be distinguished between each other. And for some reason you seems to be allowing people to change it

Having do-not-track (regardless static/dynamic) disabled is better than having it enabled because the general mass of all phones has it disabled and by making your set of parameters closer to the general public, you increase the chance of collisions.

Mozilla did it perfectly. They have it static -- user can't change it at all, because they removed the checkbox, and they made it disabled to make their users closer to average browser's parameters

thestinger:
> Fingerprint is hash sum of all parameters.

That's a very naive way to do it.

> Some of them care about JITness, some of them don't.

Vanadium makes changes to improve privacy and security which can be detected. That was an example. Vanadium does not blend in with non-hardened Chromium and Chromium itself doesn't blend in with Chrome. There are easy ways to differentiate them.

> Having static do-not-track (regardless if it's on/off) is better than having user-changeable do-not-track because you reduce the chance of individuals using vanadium to be distinguished between each other. And for some reason you seems to be allowing people to change it

Browsers have a whole bunch of site-facing configuration. This is one boolean, similar to dark mode being on or off. It's set enabled by default in Vanadium so that people don't change it from the default and blend in with other Vanadium users.

> Mozilla did it perfectly. They have it static -- user can't change it at all, because they removed the checkbox, and they made it disabled to make their users closer to average browser's parameters

They're the ones who came up with this and https://support.mozilla.org/en-US/kb/global-privacy-control was added as a more complex replacement for it. They didn't reduce fingerprinting.

PR #1127, opened 2025-02-26
Update attestation-compatibility-guide.html

groetzner-net:
Add Meteogram App

### DIFF
--- a/static/articles/attestation-compatibility-guide.html
+++ b/static/articles/attestation-compatibility-guide.html
@@ -152,7 +152,7 @@
                     <li><a href="https://play.google.com/store/apps/details?id=com.yuh">Yuh</a></li>
                     <li><a href="https://play.google.com/store/apps/details?id=de.tk.tkapp">TK-App</a></li>
                     <li><a href="https://play.google.com/store/apps/details?id=it.pagopa.io.app">IO</a> (Italian government app which uses it for the digital wallet feature)</li>
-                    <li><a href="https://play.google.com/store/apps/details?id=com.cloud3squared.meteogram">Meteogram Wetter Widget</li>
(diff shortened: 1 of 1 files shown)


static/articles/attestation-compatibility-guide.html
```diff
@@ -152,6 +152,7 @@ <h2><a href="#apps-banning-grapheneos">Apps banning GrapheneOS</a></h2>
                     <li><a href="https://play.google.com/store/apps/details?id=com.yuh">Yuh</a></li>
                     <li><a href="https://play.google.com/store/apps/details?id=de.tk.tkapp">TK-App</a></li>
                     <li><a href="https://play.google.com/store/apps/details?id=it.pagopa.io.app">IO</a> (Italian government app which uses it for the digital wallet feature)</li>
+                    <li><a href="https://play.google.com/store/apps/details?id=com.cloud3squared.meteogram">Meteogram Wetter Widget</li>
```
thestinger: There should be a `</a>` before the `</li>`.

thestinger (changes requested):
Can you confirm it shows the Play Integrity API notification when it refuses access?

static/articles/attestation-compatibility-guide.html
groetzner-net: closing-tag is added

groetzner-net: Yes can confirm the Play Integrity API notification. With anabled API the App is working, without it is not working

thestinger: But it works on GrapheneOS without blocking it? It should't be on this list if it's only using the basic integrity level and isn't blocking GrapheneOS when the Play Integrity API isn't blocked.

Issue #1149, opened 2025-04-02
Consider adding a "desktop site" check to the prerequisites section of the web installer


ethereal26:
I would like to report an issue I've experienced when using the web installer.

I have a Pixel Tablet with GOS already installed on it and tried to use it to install GOS on my new Pixel phone. Unfortunately, the flashing process worked fine at the beginning but after the first reboot of the device to be flashed it got stuck and didn't progress any further.

On the Discord server, I've been kindly advised to check if Vanadium was using the "Desktop site" mode for loading the web installer page. I tried to disable it and then the page showed a new "Reconnect device" button every time the device to be flashed needed to be reconnected.

As far as I know, on Pixel phones the default setting for Vanadium is not to load pages in "Desktop site", so what I've described above doesn't seem like a great issue. However, on the Pixel Tablet the default is to use "Desktop site" to load pages. This could be an issue for every user that wants to install GOS from a Pixel Tablet (and possibly other Android tablets that by default load web pages in "desktop site" mode).

I think it would be a good idea documenting this in the prerequisites section of the web installer. Something like "when running the web installer from an Android device, check you are not loading it in desktop site mode".

thestinger:
It should still work in the desktop site mode. I'm not sure why it wouldn't work in that mode.

ethereal26:
Yesterday, when I got a fast answer on the Discord server, I assumed this was a known bug/feature of the web installer. Today, reading you saying it should work fine, I've retried the flashing process, to be sure I hadn't missed anything yesterday.

I've found the same behavior. The "Reconnect device" button  is not displayed when using desktop mode. I'm attaching two screenshots to better show it. In one of them desktop site mode is enabled and in the other disabled.

![Image](https://github.com/user-attachments/assets/1af268bc-8f62-403b-825d-1ec75b0f7b62)
![Image](https://github.com/user-attachments/assets/c7670457-3e18-4ac2-ab71-29de32ef19a9)

ghost:
After viewing the code from the fastboot.ts source map included in this repo, I see:

```typescript
        if (navigator.userAgent.includes("Android")) {
            await this.waitForDisconnect();
            onReconnect();
        }
```

I have a feeling this is the problem. The desktop mode makes Chrome appear to be Linux and thus "Android" doesn't get matched, so it thinks its a desktop.

I'm not aware of a more robust way to check if desktop mode is in use, but an instruction to turn it off could probably work.

ghost:
[...]

PR #1162, opened 2025-04-25
Add desktop site prerequisite

ghost:
Should help with #1149.


thestinger:
This is a good idea but I'm going to add it with more detailed wording.

PR #1191, opened 2025-06-08
Opt-into edge-to-edge by default

cookiell546:
Chromium already supports dynamic edge-to-edge viewports. This change opts-in by default, making the gesture navigation bar (chin) invisible without needing scroll interaction.

No other changes were necessary, as no content relied on specific viewport insets.

Command used:

```
sed -i 's/<meta name="viewport" content="width=device-width, initial-scale=1"\/>/<meta name="viewport" content="width=device-width, initial-scale=1, viewport-fit=cover"\/>/g' **/*.html
```

### DIFF
--- a/static/404.html
+++ b/static/404.html
@@ -7,7 +7,7 @@
         <meta name="theme-color" content="#212121"/>
         <meta name="color-scheme" content="dark light"/>
         <meta name="msapplication-TileColor" content="#ffffff"/>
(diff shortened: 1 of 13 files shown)


thestinger:
This missed the files in `/articles/`, `/history/` and `/install/` so I'm going to modify it to extend it to those.

PR #1213, opened 2025-07-14
faq: Add note about IMS domains

jpds:


thestinger:
MMS is a separate thing.

thestinger:
MMS still needs to be listed. I think the approach to changing this needs to be adjusted. Not all of it involves IMS so you can't change the section to only talk about IMS. You could mention which ones are done via IMS.

PR #1229, opened 2025-07-30
Improve error handling in the web installer

cillyvms:
Some bugs in the web installer were causing errors to be handled badly:
1. Sometimes non-error object were thrown, which resulted in "Error: undefined" being displayed, as the message was formatted with `Error: ${error.message}`.
2. XMLHttpRequest code assumed that non-200 status codes would result in onerror being called, but that's not the case. I added an explicit check for status codes in onload and replaced the message in onerror with a generic "Network request failed".
3. Saving files in IndexedDB can fail because of low storage quotas, but this condition wasn't being detected, because the error is reported on the IDBTransaction's 'abort' event handler, while handlers were only registered on the IDBRequest object.

Additionally, I added a preemptive warning when the browser estimates a storage quota below 1800MiB, which is around 100MiB more than the average factory image size.

Screenshot showing the quota handling improvements:
25a78177-cf20-4ecf-91fe-b226670d03c8

### DIFF
--- a/static/js/web-install.js
+++ b/static/js/web-install.js
@@ -491,6 +491,16 @@ if ("usb" in navigator) {
     addButtonHook(Buttons.FLASH_RELEASE, flashRelease);
     addButtonHook(Buttons.LOCK_BOOTLOADER, lockBootloader);
     addButtonHook(Buttons.REMOVE_CUSTOM_KEY, eraseNonStockKey);
+
+    if (navigator.storage && navigator.storage.estimate) {
+        navigator.storage.estimate().then(estimate => {
+            // Currently factory images are ~1700MiB
+            // Show a warning if the estimated space is below 1800MiB
+            if (estimate.quota !== 0 && estimate.quota < 1024*1024*1800) {
+                document.getElementById("quota-warning-text").hidden = false;
+            }
+        });
+    }
 } else {
     console.log("WebUSB unavailable");
 }
--- a/static/js/web-install.js
+++ b/static/js/web-install.js
@@ -95,13 +95,27 @@ class BlobStore {
             request.oncomplete = () => {
                 resolve(request.result);
             };
-            request.onerror = (event) => {
-                reject(event);
+            request.onerror = () => {
+                reject(request.error);
             };
 
             if (onUpgrade !== null) {
                 request.onupgradeneeded = onUpgrade;
             }
+        });
+    }
+
+    async _wrapTransaction(transaction) {
+        return new Promise((resolve, reject) => {
+            transaction.oncomplete = () => {
+                resolve(transaction.result);
+            };
+            transaction.onerror = () => {
+                reject(transaction.error);
+            };
+            transaction.onabort = () => {
+                reject(transaction.error);
+            };
         });
     }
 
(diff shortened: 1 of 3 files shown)


thestinger:
We'll try to prioritize reviewing this but we're currently focused on getting prepared for Android 16 QPR1 due to the removal of Pixel device trees.

static/js/web-install.js
```diff
@@ -99,6 +105,20 @@ class BlobStore {
         });
     }
 
+    async _wrapTransaction(transaction) {
+        return new Promise((resolve, reject) => {
+            transaction.oncomplete = () => {
+                resolve(transaction.result);
```
inthewaves, line 111: It doesn't seem like an `IDBTransaction` has a `result` property: https://developer.mozilla.org/en-US/docs/Web/API/IDBTransaction#instance_properties. 

`_wrapTransaction` is only used for error handling and the `Promise` value isn't actually used, so I think we could get away with just `resolve()`

static/js/web-install.js
```diff
@@ -460,6 +491,16 @@ if ("usb" in navigator) {
     addButtonHook(Buttons.FLASH_RELEASE, flashRelease);
     addButtonHook(Buttons.LOCK_BOOTLOADER, lockBootloader);
     addButtonHook(Buttons.REMOVE_CUSTOM_KEY, eraseNonStockKey);
+
+    if (navigator.storage && navigator.storage.estimate) {
+        navigator.storage.estimate().then(estimate => {
+            // Currently factory images are ~1700MiB
+            // Show a warning if the estimated space is below 1800MiB
+            if (estimate.quota !== 0 && estimate.quota < 1024*1024*1800) {
```
inthewaves, line 499: The web installer page gives these explicit prerequisites, so I think this check should at least be consistent with those values: 

> You need a computer for running the web installer with at least 2GB of free memory available and 32GB of free storage space

I _think_ 32GB of free storage space is an overestimate, but we should at least set the threshold to be the specified memory requirement (2GB) to be consistent.

inthewaves (changes requested):

static/js/web-install.js
thestinger: The free space it's checking for is much higher than what it needs because the browser only allows using a small portion of that overall free space. It doesn't allow 1 site to consume all free space. That's why we have 32GB, which may or may not be enough depending on the formula it uses on that device. For example, on a 4TB drive with 16GB remaining, it may not actually allow us to use 2GB.

thestinger: I've adjusted it to 2000MiB rather than 1800MiB to have a bit more leeway so we don't need to carefully watch it.

static/js/web-install.js
cillyvms: That seems right, it must've been an oversight on my part, `resolve()` alone should work fine. Likewise `oncomplete` could be dropped from `_wrapReq` since `IDBRequest` only has a `success` event. I could make a PR, but this should be harmless, so I'm not sure if that's desired.

thestinger: It makes sense to make a PR for it.

PR #1244, opened 2025-09-01
added incompatible app

rot8:
Sources for incompatibiliy:
- https://www.agov.admin.ch/en/operating-system
- play store reviews
- tested on pixel 6 pro


thestinger:
Does it trigger the Play Integrity API notice?

rot8:
No, other than for example the SwissID app, I get no message that the Play Integrity API is used. 

agov

thestinger:
@rot8 Try disabling secure spawning and rebooting. Turn off exploit protection compatibility mode for the app.

rot8:
@thestinger Thank you, I have tried that now. However, the app displays the same screen as before.

Issue #961, opened 2024-06-26
explain it more clearer about Duress PIN


oppressor1761:
It should be explained clearly that Duress pin does not work when device password (not PIN) is requested.

thestinger:
It's a duress PIN **and** password feature. Duress PIN is for PIN entry and duress password is for password entry.

Ganwtrs:
https://github.com/GrapheneOS/grapheneos.org/pull/1246
I believe this issue should be marked as completed

PR #1246, opened 2025-09-03
Elaborate on Duress PIN/Password

Ganwtrs:
Closes Issue #961


thestinger:
Actually going to reword this and move it.

Issue #1108, opened 2025-01-29
Improve page design


ghost:
I appreciate the detailed explanations on the GrapheneOS website, but it just has too much text on each pages. Rather than explaining every single details in one page, it would be better for both newcomers and users to see a concise overview with some improvements in the page layout and UI/UX. I honestly don't think many people will read all those text... Also the 'Install GrapheneOS' on the homepage should direct to the Web installer rather than going to the install page explaining what is what.

iamavolk:
In addition, the "features" text needs do be completely revised/rewritten. It is just a prime example of bad writing, infested with run-on sentences, incoherent thoughts, and convoluted phrases. Which is kind of sad, the project itself is great.

janbaykara:
Some of us had the thought of adding a step-by-step interactive guide, to make it easy as possible for the different permutations.

Would a PR for that be accepted?

Ganwtrs:
> Rather than explaining every single details in one page, it would be better for both newcomers and users to see a concise overview

Isn't there already a consise overview in the About section of the GrapheneOS homepage?

PR #1254, opened 2025-09-17
Make the default install page the web install instructions

Ganwtrs:
The Install Index is pretty redundant, as the pages for both methods of install already link to each other. Removal of the index would lessen confusion for new users.

Addresses one of the requests for Issue #1108


thestinger:
I don't think the index page causes confusion and don't really want to remove this structure. In the future there might also be an app for performing the install from an Android device.

PR #1261, opened 2025-09-27
Mention support for DrumBrake on Vanadium

Ganwtrs:


thestinger:
This should be reordered so that it's underneath the main JIT entry.

Issue #1265, opened 2025-10-04
Add information about security preview releases


An-anonymous-coder:
The features or usage pages should have information about the security preview releases. It should also specify, for example, if the [2025092700 stable SP] release is prioritized above the [2025100300 beta non-SP] release, or if that is considered a downgrade.

PR #1283, opened 2025-11-08
Security preview releases

Ganwtrs:
Partially addresses #1265 , but I don’t know the answer to what @An-anonymous-coder proposed

> It should also specify, for example, if the [2025092700 stable SP] release is prioritized above the [2025100300 beta non-SP] release, or if that is considered a downgrade.


thestinger:
Please resubmit this with a cleaned up history based on the latest main branch. It should have 1 commit if it's making 1 change. You can fix this by rebasing it but if you don't know how to do that, you can make the commit again.

thestinger:
You can rebase it here without opening a new pull request if you want to do it that way, which is what I would prefer, but I assume the reason it's this way is because you're using the GitHub web interface which lacks support for making a pull request in a form that can be accepted i.e. not with a messy history.

thestinger:
@Ganwtrs Can you just reopen it? It looks like the change got dropped.

PR #1285, opened 2025-11-10
specify the classic version of yarn

cookiell546:
Yarn Modern (v2+) formats its lockfile differently, and would mutate the lockfile of adevtool. This issue became apparent while supporting a user in the developement room.

Most distributions package Yarn Classic anyway.


thestinger:
Why list both yarn and yarn classic?

cookiell546:
My fault while rebasing and resolving the conflict.

PR #1307, opened 2025-11-26
Update phone display to Pixel 10 Pro XL

An-anonymous-coder:

PR #1308, opened 2025-11-27
Fix HTML formatting

An-anonymous-coder:
I fixed the HTML formatting for every file. Because I was stupid and don't know how forks or branches or pull requests work, this is a branch of https://github.com/GrapheneOS/grapheneos.org/pull/1307. I fixed a couple grammar errors on the front page and the features page. The rest of the content is unchanged.


thestinger:
I don't really understand what you're trying to fix and it's too much mixed together to decipher it.

thestinger:
Please submit fixes in small pull requests with only those fixes. There should be clear reasoning for doing it. The way you're changing the HTML formatting isn't fixing it and is going against our HTML style.

An-anonymous-coder:
I'm tired

An-anonymous-coder:
Where are you getting your HTML style from?

An-anonymous-coder:
Alternatively, is there a contributing guideline that lists that out?

An-anonymous-coder:
My main goal was to fix the various grammar issues across the website, but my HTML formatter doesn't format in the same way, so I decided to first make a PR to align the formatting to match. I also can't easily or cleanly make changes, since I can't make a second fork without interfering with the main branch of the other PR. 

I am extremely bad at pull requests.

thestinger:
You shouldn't use the main branch to implement anything for pull requests. I don't know how to help now beyond that you can make a new clone and start over there with proper feature branches.

PR #1321, opened 2025-12-02
Maintain a single source for the verified boot key hashes

rdevshp:
Currently there are two copies of the verified boot key hashes. This commit moves it to a template that is included in the two files.

### DIFF
--- a/templates/verified_boot_key_hash.html
+++ b/templates/verified_boot_key_hash.html
@@ -1,0 +1,43 @@
+<section id="verified-boot-key-hash">
+    <h3><a href="#verified-boot-key-hash">Verified boot key hash</a></h3>
+
+    <p>When loading an alternate OS, the device shows a yellow notice on boot
+    with the ID of the alternate OS based on the sha256 of the verified boot
+    public key. 4th and 5th generation Pixels only show the first 32 bits of
+    the hash so you can't use this approach. 6th generation Pixels onwards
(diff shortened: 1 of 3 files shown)


templates/verified_boot_key_hash.html
```diff
@@ -0,0 +1,43 @@
...
+    <p>Checking this is useful after installation, but you don't need to check
+    it manually for verified boot to work. The verified boot public key
+    flashed to the secure element can only be changed when the device is
+    unlocked. Unlocking the device performs the same wiping of the secure
+    element as a factory reset and prevents data from being recovered even if
+    the SSD was cloned and your passphrase(s) are obtained because the
+    encryption keys can no longer be derived anymore. The verified boot key is
+    also one of the inputs for deriving the encryption keys in addition to the
+    user's lock method(s) and random token(s) on the secure element.</p>
+</section>
```
thestinger: Missing a trailing newline in this file per standard Unix text file format.

rdevshp: Fixed.

rdevshp:
Anything else I need to do to get this merged? Thanks!

jermanuts:
@thestinger when will this be merged as CLI installation guide is currently missing [Verified boot key hash](https://grapheneos.org/install/cli#verified-boot-key-hash) for Pixel 10 devices?

PR #1329, opened 2025-12-09
Add age encryption details to contact page

Ganwtrs:
GrapheneOS has an age public key

https://primal.net/e/nevent1qqs8x6x0y29mylmrakd48zcjg8h670r65dfwlpq2l0896xv6uvyg63c566u40


thestinger:
Can you remove the link here too? That only exists because the security.txt standard is stupid.

thestinger:
It should also be wrapped with <code></code> and there should be a period at the end of the paragraph after that.

PR #1330, opened 2025-12-09
add PosteID to list of apps banning GrapheneOS

jacopotediosi:
The PosteID app, owned by Poste Italiane - the Italian national postal operator - has recently adopted the Play Integrity API, which now prevents the application from running on GrapheneOS and results in an error message at startup.

The developer email listed on the Play Store, `assistenza_posteapp@poste.it`, is not functional. This morning I attempted to file a complaint via `servizio.clienti@posteitaliane.it`; I will update you if I receive a response.

Refs:
- https://github.com/GrapheneOS/os-issue-tracker/issues/6646
- https://discuss.grapheneos.org/d/28060-posteid-spid-provider-on-grapheneos/21

### DIFF
--- a/static/articles/attestation-compatibility-guide.html
+++ b/static/articles/attestation-compatibility-guide.html
@@ -161,6 +161,7 @@
                     <li><a href="https://play.google.com/store/apps/details?id=com.swisssign.swissid.mobile" rel="nofollow">SwissID</a></li>
                     <li><a href="https://play.google.com/store/apps/details?id=de.tk.tkapp" rel="nofollow">TK-App</a> (German health insurance app which uses it for fingerprint login)</li>
                     <li><a href="https://play.google.com/store/apps/details?id=it.pagopa.io.app" rel="nofollow">IO</a> (Italian government app which uses it for the digital wallet feature)</li>
+                    <li><a href="https://play.google.com/store/apps/details?id=posteitaliane.posteapp.appposteid" rel="nofollow">PosteID (Italian postal service’s app used to access the national digital identity system "SPID")</a></li>
                     <li><a href="https://play.google.com/store/apps/details?id=sg.ndi.sp" rel="nofollow">Singpass</a></li>
                 </ul>
 


thestinger:
> The developer email listed on the Play Store, assistenza_posteapp@poste.it, is not functional

This should be reported as a Play Store policy violation.

jacopotediosi:
> This should be reported as a Play Store policy violation

Just done:

image

orazioedoardo:
This is definitely not true, I’ve had two independent confirmations today that the app works fine with GrapheneOS including actually authenticating to a public service. The app was simply installed from Play Store with preexisting Play Services, without relaxing exploit protections.

It’s possible that they are using Play Integrity to block first login on the app and not subsequent usage, but recent Play Store show many users having login errors so it may not be Play Integrity at all.

It would be quite atrocious if they actually end up using it though since it is the main identity provider in Italy and the fact that the app is just a TOTP wrapper. Maybe if they are not using certificate pinning the TOTP seed can be extracted to work in a third party app.

jacopotediosi:
I've a friend which reversed their APIs and confirmed they use Play Integrity in their latest version. I just asked him to post his findings here.

thestinger:
Using it is different from blocking GrapheneOS users with it. Do they block people logging in?

jacopotediosi:
When opening the app immediately after a fresh installation, before even reaching the login screen, the following error appears and prevents any further interaction, at the same time as the GrapheneOS popup indicates that the app has checked Play Integrity.

Screenshot_20251224-080156

If the Play Integrity request is blocked, the error message changes to the following:

Screenshot_20251224-080218

~I am not entirely sure I remember this correctly, which is why I asked my friend to write here, but it seems to me that he told me that the new version of the APIs the app is migrating to uses server-side attestation also for issuing the TOTP seed.~

thestinger:
Okay, going to leave it on the list for now.

orazioedoardo:
That's very disappointing, shame on them. I didn't expect them to do that with the identity provider app before the banking app. As far as I know there's no requirement from AgID to do this. Maybe the Play Store reviews are from users with broken attestation.

orazioedoardo:
> I've a friend which reversed their APIs and confirmed they use Play Integrity in their latest version.

The app allegedly supports as far as Android 5, which doesn't support hardware attestation so either it is broken on those devices, or they have an alternative API and thus a downgrade attack path.

lakor64:
Hello, I can confirm that on the latest PosteID version (4.11.8 (17608)) at the time I started looking at this), the application does require Play Integrity in order to function.
The integrity check is done during the application initial registration to Poste server, before any login can even be attempted.

Their new endpoint has started to enforce Play Integrity recently (hence why it used to work before) and, as the integrity check is done only during the initial registration, anyone who has already obtained a token from Poste server will not experience any issue.
```
{
   "integrityCheck":"enabled",
   "iss":"jod-app-registry",
   "publicKey":"",
   "exp":1765475069,
   "registerToken":"[...]",
   "iat":1765475009
}
```
Whenever the server answers the request with "integrityCheck": "enabled" it forces the app the send the integrity token to be verified back to the server, whenever this request fails the server sends an error with the "AI-01" error
My phone which only passes PlayIntegrity basic is affected by the same issue.
This is an example request that the client sends and waits for a validation result from the server:
{
  "sub": "register",
  "integrity": {
    "integrityToken": "CugBARCnMGt_81Gg[....]" <-- here the play integrity token
  },
  "nbf": 1766594575,
  "osVersion": "5.0",
  "iss": "app-posteid-v3",
  "rooted": "false",
  "deviceModel": "[...]",
  "exp": 1766594635,
  "registerToken": "[...]",
  "iat": 1766594575,
  "jti": "[..]"
}
I don't know when the server started to mandatory request integrity and based from my findings the server always answers enforced even when reporting Android 5.0 (my suspect is that they simply killed the compatibility for it)

orazioedoardo:
@lakor64 Does it check for `MEETS_DEVICE_INTEGRITY` or `MEETS_STRONG_INTEGRITY` ?

jacopotediosi:
> As far as I know there's no requirement from AgID to do this.

Maybe they copied from the European Digital Identity, which until July 2025 listed in its requirements the use of Play Integrity (see commit: https://github.com/eu-digital-identity-wallet/eudi-app-android-wallet-ui/commit/9367289f3da1abd7f2cff0e7e05d67a95f4fdf60).

The requirement was removed following this discussion: https://github.com/eu-digital-identity-wallet/eudi-app-android-wallet-ui/issues/287#issuecomment-3124658054.

I referenced these sources in the email sent to Poste Italiane at the same time as this PR, in which I requested the removal of Play Integrity, but I have not yet received a response.

lakor64:
> @lakor64 Does it check for `MEETS_DEVICE_INTEGRITY` or `MEETS_STRONG_INTEGRITY` ?

I belive it checks for STRONG_INTEGRITY, but I don't have a mobile phone that passes DEVICE_INTEGRITY and fails STRONG_INTEGRITY to check. (My phone does pass only MEETS_BASIC_INTEGRITY)
Anyway grapheneOS by default shouldn't pass DEVICE_INTEGRITY as well.

orazioedoardo:
> Maybe they copied from the European Digital Identity, which until July 2025 listed in its requirements the use of Play Integrity

I don't think that applies to SPID though. It's possible Poste Italiane is doing this in the aftermath of (unrelated) identity thefts via SPID. They are trying to implement every random security check just in case to claim they did everything to protect the user from fraud and can't be sued.

jacopotediosi:
Well, apparently according to Google it is not a violation of policy to have contacts in the Play Store listing not working:
image

AlessandroBonomo28:
> > @orazioedoardo  @lakor64 Does it check for `MEETS_DEVICE_INTEGRITY` or `MEETS_STRONG_INTEGRITY` ?
> 
> I belive it checks for STRONG_INTEGRITY, but I don't have a mobile phone that passes DEVICE_INTEGRITY and fails STRONG_INTEGRITY to check. (My phone does pass only MEETS_BASIC_INTEGRITY) Anyway grapheneOS by default shouldn't pass DEVICE_INTEGRITY as well.

As today, it only checks for DEVICE_INTEGRITY without STRONG_INTEGRITY and BASIC_INTEGRITY.
See here: https://discuss.grapheneos.org/d/28060-posteid-spid-provider-on-grapheneos/45

PR #1334, opened 2025-12-10
Fix typo

nazar-pc:
There was one typo in the date, then another. But this is actually the correct one.


nazar-pc:
Okay, it is fixed now. Though force-pushing into `main` is... not great.

thestinger:
I didn't see that this was submitted. We don't treat the website repository as something that people might be depending on having immutable history for downstream projects. There's a **lot** of simply amending recent commits which goes on.

PR #1339, opened 2025-12-12
Mention effects of auto-confirm in duress PIN behavior description

Ganwtrs:


thestinger:
I think it very likely also prevents it if it's shorter, not just longer. I don't feel like trying right now.

PR #1345, opened 2025-12-18
static/install/web: clarify device requirements

eclairevoyant:
This clarifies that a _computer_ in the PC sense isn't necessary.
(I assume this sentence was copied from the CLI install instructions without sufficient modification, originally.)


thestinger:
It's not because it was copied from somewhere but it might be clearer to people who don't consider small computers to be computers.

PR #1349, opened 2025-12-22
update best development devices

Ganwtrs:
<details>The best GrapheneOS development devices are no longer Pixels series 🫲6 7🫱</details>


thestinger:
It should include Pixel 10 in this.

PR #1364, opened 2026-01-02
add TK-Ident & TK-Doc to list of apps banning GrapheneOS

felschr:
For context, I originally suggested adding these apps here: https://bsky.app/profile/felschr.com/post/3mbhn6xphok2g

I also updated info for TK-App including requirements how to at get it partially working on GrapheneOS.

Let me know if the nested list & footnotes work for you or if I should change something.

### DIFF
--- a/static/articles/attestation-compatibility-guide.html
+++ b/static/articles/attestation-compatibility-guide.html
@@ -159,7 +159,9 @@
                     <li><a href="https://play.google.com/store/apps/details?id=com.mcdonalds.mobileapp" rel="nofollow">McDonald's</a> (International app used for many but not all countries not including the US)</li>
                     <li><a href="https://play.google.com/store/apps/details?id=com.ridedott.rider" rel="nofollow">Dott</a></li>
                     <li><a href="https://play.google.com/store/apps/details?id=com.swisssign.swissid.mobile" rel="nofollow">SwissID</a></li>
-                    <li><a href="https://play.google.com/store/apps/details?id=de.tk.tkapp" rel="nofollow">TK-App</a> (German health insurance app which uses it for fingerprint login)</li>
+                    <li><a href="https://play.google.com/store/apps/details?id=de.tk.tkapp" rel="nofollow">TK-App</a> (Blocks access to TK-Safe, TK-GesundheitsMessenger, fingerprint login)</li>
+                    <li><a href="https://play.google.com/store/apps/details?id=de.telearzt.tkchat" rel="nofollow">TK-Doc</a></li>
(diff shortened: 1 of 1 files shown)


static/articles/attestation-compatibility-guide.html
```diff
@@ -175,6 +181,12 @@ <h2><a href="#apps-banning-grapheneos">Apps banning GrapheneOS</a></h2>
                 with compatibility or security issues. The only reason they aren't permitting it is
                 because we do not license Google Mobile Services (GMS) and these apps are enforcing
                 Google's business interests rather than security.</p>
+
+                <p>
+                  <sup>1</sup>Requires Google Play Services<br/>
```
thestinger: We don't want to include whether an app depends on Play services here. It's definitely out-of-scope. We want to stick to whether apps are banning GrapheneOS for this section. We also don't want to have a section like this explaining markers at the bottom.

static/articles/attestation-compatibility-guide.html
```diff
@@ -159,7 +159,13 @@ <h2><a href="#apps-banning-grapheneos">Apps banning GrapheneOS</a></h2>
                     <li><a href="https://play.google.com/store/apps/details?id=com.mcdonalds.mobileapp" rel="nofollow">McDonald's</a> (International app used for many but not all countries not including the US)</li>
                     <li><a href="https://play.google.com/store/apps/details?id=com.ridedott.rider" rel="nofollow">Dott</a></li>
                     <li><a href="https://play.google.com/store/apps/details?id=com.swisssign.swissid.mobile" rel="nofollow">SwissID</a></li>
-                    <li><a href="https://play.google.com/store/apps/details?id=de.tk.tkapp" rel="nofollow">TK-App</a> (German health insurance app which uses it for fingerprint login)</li>
+                    <li>Various apps by Techniker Krankenkasse (German public health insurance organization)
```
thestinger: These should just be at the top level instead of grouped together. It will get too subjective deciding which apps should be grouped together especially if the developers change.

thestinger (changes requested):

static/articles/attestation-compatibility-guide.html
felschr: Should I keep the "partially works" remark & list of blocked features for TK-App just without the footnotes or remove it altogether?
Without any of those 3 requirements, the app fails directly on launch and is entirely unusable.

thestinger: The only thing relevant to this section is whether apps ban using it at all via the Play Integrity API or ban using specific parts. If only specific parts are banned, that should be noted. Whether apps depend on Google Play isn't relevant to this section and we aren't going to be listing millions of apps depending on Google Play.

felschr: Alright, updated it accordingly.

PR #1380, opened 2026-01-23
Specify how features like MTE are only supported on Pixels 8th gen and above

Ganwtrs:
For the FAQ, now that Pixel 10 support is production ready, it should be mentioned how not just Pixels 8th and 9th generation have 7 year support and MTE. 

For the features section, I specified that only Pixels 8th gen and above have support for MTE.

### DIFF
--- a/static/features.html
+++ b/static/features.html
@@ -355,11 +355,11 @@
                                     allocations to block C string overflows, absorb small overflows
                                     and detect linear overflows or other heap corruption when the
                                     canary value is checked (primarily on free)</li>
-                                    <li>Hardware memory tagging for slab allocations (128k and
-                                    below) providing probabilistic detection of all use-after-free
-                                    and inter-object overflows along with deterministic detection
-                                    of all small/linear overflows and use-after-free until it has
-                                    been reused once and gone through the quarantines twice</li>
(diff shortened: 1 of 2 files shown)


static/features.html
```diff
@@ -1076,7 +1076,8 @@ <h3><a href="#vanadium">Vanadium: hardened WebView and default browser</a></h3>
                     <p>Some of the features added compared to standard mobile Chromium:</p>
 
                     <ul>
-                        <li>Hardware memory tagging (MTE) enabled for the main allocator</li>
+                        <li>Hardware memory tagging (MTE) enabled for the main allocator (supported
```
thestinger, line 1079: Pixels should be capitalized here

thestinger (changes requested):
It's worth noting we're going to have non-Pixel devices supported in 2027 so it would be good to prepare for that. Not sure exactly how to approach it.

Ganwtrs:
> It's worth noting we're going to have non-Pixel devices supported in 2027 so it would be good to prepare for that. Not sure exactly how to approach it.

Maybe a comparison table between supported devices, like how [secureblue compares desktop environments](https://secureblue.dev/images#security-recommendation)?

PR #1392, opened 2026-02-03
add EasyPark to list of apps banning GrapheneOS

jacopotediosi:
[EasyPark](https://play.google.com/store/apps/details?id=net.easypark.android) is a widely used parking payment app in Europe, particularly popular in Northern Italy and supported by nearly all the municipalities.

The Android app refuses to send the SMS OTP if the device fails Play Integrity checks.

I contacted appsupport@easypark.net and support@easypark.net requesting GrapheneOS support and linking the [Attestation Compatibility Guide](https://grapheneos.org/articles/attestation-compatibility-guide).

After several AI-generated responses, I received a human reply confirming that the app intentionally blocks ROMs that fail Play Integrity, including GrapheneOS, citing "security reasons":

image

Fun fact: obviously this is not a valid security measure, since it's possible to authenticate by requesting an OTP from the app (which will never arrive), then requesting one from the website and using it to authenticate in the app as if it were the code from the first request. However, I would consider this bypass as a vulnerability in EasyPark, so I would still recommend adding the app to the list of those that ban GrapheneOS.

### DIFF
--- a/static/articles/attestation-compatibility-guide.html
+++ b/static/articles/attestation-compatibility-guide.html
@@ -163,6 +163,7 @@
                     <li><a href="https://play.google.com/store/apps/details?id=de.tk.apps.android.tkident" rel="nofollow">TK-Ident</a></li>
                     <li><a href="https://play.google.com/store/apps/details?id=de.tk.tkapp" rel="nofollow">TK-App</a> (Blocks access to TK-Safe, TK-GesundheitsMessenger, fingerprint login)</li>
                     <li><a href="https://play.google.com/store/apps/details?id=it.pagopa.io.app" rel="nofollow">IO</a> (Italian government app which uses it to gate access to the digital wallet feature)</li>
+                    <li><a href="https://play.google.com/store/apps/details?id=net.easypark.android" rel="nofollow">EasyPark</a></li>
                     <li><a href="https://play.google.com/store/apps/details?id=posteitaliane.posteapp.appposteid" rel="nofollow">PosteID</a> (Italian postal service’s app used to access the national digital identity system "SPID")</li>
                     <li><a href="https://play.google.com/store/apps/details?id=sg.ndi.sp" rel="nofollow">Singpass</a></li>
                 </ul>


ron-png:
I could login with minor hiccups: 
- Had to give play services SMS permission.
- I received the SMS (I am using google messages with RCS enabled)
  - i think this is important information because I had a problem with telegram verification code when using the stock SMS app.

jacopotediosi:
They confirmed via mail that they ban Graphene on purpose, so this is not an opinion

ron-png:
> They confirmed via mail that they ban Graphene on purpose, so this is not an opinion

They only say that they do not plan on supporting custom ROMs specifically, not that they are targeting GrapheneOS.
It's the same with Revolut. It needed a workaround to get the app running on GrapheneOS as well. But it doesn't mean that they specifically ban it.

ron-png:
@thestinger if you have the time, I'd like to hear your opinion on this matter. 
As far as I understand it, most Apps nowadays use the Play integrity API, but not always the deepest level of it (which is the root of the problem). 
In this case, easypark is clearly not using the App-breaking level, as I could log in with some tinkering.

So my question is: 
At what point can we say that an App developer bans GrapheneOS? Do we call it banned once the developer answers implying it or once it doesn't work in practice?

jacopotediosi:
I just tested and they still don't send the sms to complete the login if you fail device integrity

ron-png:
> I just tested and they still don't send the sms to complete the login it you fail device integrity

So why did it work for me then?

jacopotediosi:
Please don't make me set up a proxy again to prove it by showing the network traffic, I already did it a few months ago

jacopotediosi:
> So why did it work for me then?

Maybe you are using a different build, I downloaded mine from the Google Play Store, with Country set to Italy

ron-png:
> > So why did it work for me then?
> 
> Maybe you are using a different build, I downloaded mine from the Google Play Store, with Country set to Italy

That's interesting. I have a German Play store  account.

What SMS app are you using? 

Screenshot_20260405-130527

lakor64:
Last version of Easypark downloaded from Google Play Store as of 14-05-2026

Country set to: Italy
The application opens and remains stuck at SMS verification for account creation, basically I insert my phone number, the application shows me the UI to insert the SMS code but no SMS ever arrives.

SMS carrier and SMS application do work correctly, I think it's an application issue which sounds compatible with what the OP did, I have not done further research but the app does not work for me.

thestinger:
Did you try the per-app compatibility mode?

PR #1399, opened 2026-02-05
Move Dialer link to standalone apps

oscar-labit:
I think the Dialer should be moved to the standalone apps section. There have been a number of [PRs ](https://github.com/GrapheneOS/platform_packages_apps_Dialer/pulls?q=is%3Apr) to improve the dialer beyond the AOSP version, including implementing call recording, improved call quality, bug fixes, theme updates, and Gradle migration attempts. Additionally, there are attempts to migrate the dialer to [material design 3](https://github.com/GrapheneOS/platform_packages_apps_Dialer/pull/54), further separating the app from the AOSP. 

By moving the Dialer app to the standalone app section, community involvement for the app should increase.

### DIFF
--- a/static/source.html
+++ b/static/source.html
@@ -104,7 +104,6 @@
                     <li><a href="https://github.com/GrapheneOS/platform_packages_apps_CellBroadcastReceiver">platform_packages_apps_CellBroadcastReceiver</a></li>
                     <li><a href="https://github.com/GrapheneOS/platform_packages_apps_Contacts">platform_packages_apps_Contacts</a></li>
                     <li><a href="https://github.com/GrapheneOS/platform_packages_apps_DeskClock">platform_packages_apps_DeskClock</a></li>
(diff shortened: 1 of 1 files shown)


FSaurenbach:
AFAIK standalone means that it can be built without the aosp source code with Gradle or smth, so this doesn't make sense to merge until it has been migrated to gradle or something else
Also no expert on this but can you even build an apk that request privileged permissions without the full aosp source (it has to be signed by platform cert right?)

oscar-labit:
https://github.com/GrapheneOS/platform_packages_apps_Dialer/pull/53

That's what I spend December working on. It's not 100% perfect yet, but I've been running this Gradle build on my Pixel 8 Pro for about 2 months now with limited issues. Unfortunately, I don't have time to finish it, and the PRs take months to review because nobody's monitoring the repository.

PR #1401, opened 2026-02-06
Remove TrustCor's distrust from features list

FSaurenbach:
update distrusted system CAs to Chunghwa Telecom and Netlock


thestinger:
Instead of removing this, it should be updated to say Chunghwa Telecom and Netlock are removed.

FSaurenbach:
@thestinger done

PR #1418, opened 2026-02-13
Revise GrapheneOS Camera functionality description

Ganwtrs:
I do agree that GrapheneOS Camera is far better than any Android Camera apps I am aware of, but I feel that the description could be reworded to be a bit more humble

### DIFF
--- a/static/usage.html
+++ b/static/usage.html
@@ -657,11 +657,10 @@
                 <p>GrapheneOS has the same camera capabilities and quality as the stock OS. It
                 will match the stock OS when comparing the same app on each OS. GrapheneOS uses
                 our own modern Camera app rather than the standard AOSP Camera app. GrapheneOS
-                Camera is far better than any of the portable open source camera alternatives and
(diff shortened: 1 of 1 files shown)


thestinger:
You're changing the meaning of what was said beyond what's described and in a way that's not wanted.

Ganwtrs:
On the main page, it could theoretically go something like “GrapheneOS is far better than any other consumer operating system for privacy and security (in certain other contexts that would be specified)”, but that would be less humble IMO compared to the existing “GrapheneOS improves the privacy and security of the OS from the bottom up” even if the meaning ends up being different. Readers can understand that GrapheneOS has comprehensive improvements in both examples, and the rest of the website provides technical explanation for these improvements.

In the case of the Camera description scenario, even though the meaning is changed, I believe it would be more in line with the way the GrapheneOS website describes improvements in GrapheneOS related services.

thestinger:
You changed it beyond the way you said you did. Not really interested in debate about it. It can be changed but not like that because you're changing the meaning in a way that's not what is intended to be said.

PR #1419, opened 2026-02-13
Revise camera functionality description v2

Ganwtrs:
Please let me know if anything falls into the intended definition of “far better” that goes beyond privacy, security, and functionality improvements.


thestinger:
No, still don't agree with this.

PR #1429, opened 2026-02-15
For contact page, revise slashes and mention how developers are active on Discord too

Ganwtrs:


thestinger:
Developers are active on both Matrix and Discord so it should say that instead.

PR #1433, opened 2026-02-18
copy udev rules installation commands from cli.html to web.html

MoellJ:
* As the web guide is meant for less tech-savy users, including explicit commands makes sense


thestinger:
It doesn't include this because the idea is that people would install it via a GUI for Ubuntu / Debian rather than CLI if they're not technical.

PR #1472, opened 2026-03-30
Added i18n support for russian, french, spanish, german languages

olekasndrsvl:
Improve localization workflow, translations, and i18n test coverage

- localize web installer UI messages and error handling through the existing gettext pipeline
- add DOMException-based error message mapping for friendlier translated WebUSB errors
- update Russian, German, Spanish, and French translations for shared navigation, homepage content, and web installer strings
- add translator comments in source templates to provide context for ambiguous strings
- add locale/report_untranslated.py to report untranslated entries by language and source file
- add locale/FOR-TRANSLATORS.MD with translation workflow and reporting instructions
- add pytest and Playwright-based localization and interaction tests under test/
- document how to install and run localization tests
- refresh AGENTS.md and project documentation for the current i18n and testing workflow
- normalize internal comments and developer-facing notes to English in scripts and configuration files


thestinger:
You're proposing making a huge set of changes without a proper explanation or coordination with us. We haven't talked about any of this. If you want to make deep changes, you need to work out the right way to do that with us prior to proposing it or it's going to be rejected because it's doing stuff we never said we wanted or determined was the right approach. Everything included in any part of GrapheneOS needs to be thoroughly reviewed and fully understood.

We cannot merge AI generated translations without fully understanding them and being able to fix the many mistranslated which are going to have happened. It's not easy to write documentation with wording getting all of the subtleties correct in one language let alone translating that to another language while preserving all of the original intent and subtleties present in it. An LLM is not capable of producing high quality translations for our documentation, error strings, etc. It's possible to use an LLM to assist with translations by humans but we won't be using an LLM to do the translations alone. In order to have translations, we'll need to have teams of people for each language who can quickly catch up to the latest state of things with accurate translations which have been reviewed and improved by multiple people.

Translations need to be done in separate repositories not interfering with development. There cannot be pull requests and commits for ongoing translations in the main repositories. That will severely interfere with development and the regular workflows. It has to be separate and has to degrade in a reasonable way when it falls behind including showing which parts are out-of-date so people can refer to the updated content. There's a whole lot to figure out if you want to have translations for the website.

This approach isn't going to work and needs to be entirely redone.

PR #484, opened 2022-08-12
Add Document-Policy header

Seirdy:
This header accomplishes the following:

- Forbids document.write
- Forbids document.domain
- Forbids use of profiling APIs
- Forbids popups (similar to the overly-agressive "sandbox" CSP
  directive; uplifed from the deprecated Feature-Policy header)

These are preffed off in Chromium as "experimental"; the only DP
directive currently enabled in Chromium is "force-load-at-top".

More information:

- Document-Policy explainer:
  https://github.com/wicg/document-policy/blob/main/document-policy-explainer.md

- Document-Policy specification:
  https://wicg.github.io/document-policy/

- Current directives supported in Chromium:
  https://chromium.googlesource.com/chromium/src/+/refs/heads/main/third_party/blink/renderer/core/permissions_policy/document_policy_features.json5

To try this out, go to "chrome://flags" and enable experimental web
platform features. See implementation status at
https://bugs.chromium.org/p/chromium/issues/detail?id=993790.

PR #1513, opened 2026-05-06
Add Document-Policy header

RankoR:
Replaces stale https://github.com/GrapheneOS/grapheneos.org/pull/484


thestinger:
Need to fix the static analysis check by adding it to location blocks where other headers are set since that drops all the outer add_header values. We're using nginx 1.30.0 so we could use add_header_inherit but I haven't had time to deal with integrating that and we need to be careful not to break it.

https://nginx.org/en/docs/http/ngx_http_headers_module.html#add_header_inherit

RankoR:
@thestinger should be fine now?

thestinger:
It needs to be tested with the web installer before I can merge it. We should also do the same thing for attestation.app which I can merge right away since we've written all that code ourselves.

PR #1530, opened 2026-05-17
Document VoWi-Fi runs outside VPN

de0u:
A forum user has requested that the documentation be updated to state that VoWi-Fi runs outside of VPNs: https://discuss.grapheneos.org/d/35281-grapheneos-version-2026050400-released/18


de0u:
Revision requests are welcome.

PR #1546, opened 2026-06-06
Add My Seat app

cvlc12:
Latest update provided the following changelog and made the app inop on GOS

> We’ve implemented new device integrity checks to ensure your app environment is secure, safeguarding your vehicle and personal data against potential risks.


thestinger:
Does it trigger the Play Integrity API notice?

cvlc12:
Yes, it does, and shows a message about the app not being able to run on rooted or modified phones. Blocking play integrity usage attempts changes the error message to "app is in maintenance".

thestinger:
Can you reorder it so that the package name in is alphabetical order? It should be one lower than it is.

cvlc12:
> Can you reorder it so that the package name in is alphabetical order? It should be one lower than it is.

Sure, it'll just have to wait a few days, I'm away and it's a pain from a phone.

cvlc12:
> Can you reorder it so that the package name in is alphabetical order? It should be one lower than it is.

done

PR #1554, opened 2026-06-17
replace front page device with Pixel 9 frame

thomasbuilds:
screenshot of change:

Screenshot from 2026-06-17 21-40-40


thomasbuilds:
let me know if you want phone's frame border lighter color / thinner bezels or fine as is

thestinger:
This should likely be a Pixel 10 Pro instead now.

thomasbuilds:
Pixel 9/10 basically has the same design. Also, this SVG isn't "official" - I made it to look roughly the same. Do you want this PR to be reopened?

thestinger:
Yes, can you reopen it? It's likely better than what we have now, but it needs to be reviewed.

thestinger:
I don't remember how the current Pixel 7 Pro one was made. It may have been made based on a reference provided by Google.

thomasbuilds:
I re-opened it, everything is the same, I just replaced string pixel 9 with pixel 10

PR #1559, opened 2026-06-20
build notes: add information about zswap and zram

### LINKED ISSUE #1558: Add notes about zswap and zram for building guide
Reason: this would lower the barrier of entry to contribute to GrapheneOS a lot

RAM currently is expensive... And without utilizing one of those technologies it is hard or nearly impossible to build GrapheneOS on machine with only 16 GiB of RAM or less.

References for resources which helped me on learning about zram and zswap including building AOSP on less then 16 GiB of RAM:
- https://www.merinorus.com/blog/building-android-roms-limited-ram-zram-zswap-comparison/
- https://linuxblog.io/running-out-of-ram-linux-add-zram
- https://linuxblog.io/zswap-better-than-zram

I tested that and can confirm that it IS possible build GrapheneOS with 16 GiB of RAM:
- without any swap partition on drive/disk just using zram with zstd and zram size around ~50GiB
- with zswap

flexxxxer:
Closes #1558

### DIFF
--- a/static/build.html
+++ b/static/build.html
@@ -234,7 +234,21 @@ source ~/.bashrc</pre>
                         <li>32GiB of memory or more. Link-Time Optimization (LTO) creates huge peaks
                             during linking and is mandatory for Control Flow Integrity (CFI). Linking
                             Vanadium (Chromium) and the Linux kernel with LTO + CFI are the most memory
-                            demanding tasks.</li>
+                            demanding tasks.
+                            <ul>
+                                <li>consider using zswap (require swap, preferred if you have NVME storage) 
+                                    or zram (does not require swap). It's possible to build GrapheneOS with 
+                                    16GiB of memory. Recommended parameters:
+                                    <ul>
+                                      <li><code>zswap.max_pool_percent=80</code> for zswap</li>
+                                      <li><code>zswap.compressor=zstd</code> for zswap</li>
+                                      <li><code>zram-size = 51200</code> for zram</li>
+                                      <li><code>compression-algorithm = zstd</code> for zram</li>
+                                    </ul>
+                                </li>
+                            </ul>
+                        </li>
+
                         <li>100GiB+ of additional free storage space for a typical build of the entire
                         OS for a multiarch device</li>
                     </ul>


static/build.html
```diff
@@ -234,7 +234,21 @@ <h3><a href="#build-dependencies">Build dependencies</a></h3>
                         <li>32GiB of memory or more. Link-Time Optimization (LTO) creates huge peaks
                             during linking and is mandatory for Control Flow Integrity (CFI). Linking
                             Vanadium (Chromium) and the Linux kernel with LTO + CFI are the most memory
-                            demanding tasks.</li>
+                            demanding tasks.
+                            <ul>
+                                <li>if you don't have enough memory consider using zswap (require swap partition, preferred
```
thestinger: Either zswap or zram should always be used regardless of how much memory someone has because it's very bad not having any form of swap. The Linux kernel requires having a form of swap in order for it to function properly, otherwise it can only reclaim memory by pushing code/data out of page cache which it then has to pull right back into page cache in many cases. It needs a way to flush out dirty memory that's not being used to avoid losing page cache which is being used. Userspace code needs to be loaded into page cache to run so having the page cache get fully exhausted is pretty bad.

flexxxxer: Propose to replace "if you don't have enough memory consider using zswap (require swap partition..." on "consider using zswap (require swap partition..." then

static/build.html
```diff
@@ -234,7 +234,21 @@ <h3><a href="#build-dependencies">Build dependencies</a></h3>
...
+                                    if you have NVME storage) or zram (does not requre swap partition). It's possible to build
+                                    GrapheneOS with 16GiB of memory. Recommended parameters:
+                                    <ul>
+                                      <li><code>zswap.max_pool_percent=80<code> for zswap</li>
+                                      <li><code>zswap.compressor=zstd<code> for zswap</li>
+                                      <li><code>zram-size = 51200<code> for zram</li>
+                                      <li><code>compression-algorithm = zstd<code> for zram</li>
+                                    </ul>
```
ryrona2, line 247: I would like if there was commands one could just copy-paste to enable either zswap or zram. Like a sequence of "modprobe zram; zramctl --algo zstd --size 50GiB; mkswap -U clear /dev/zram0; swapon --discard /dev/zram0" (verify these, I just wrote this from memory). So the user does not have to figure out how to do that themselves. Just a suggestion!

flexxxxer: It depends on distribution user has. As stated in build instructions:
> Arch Linux, Debian bookworm, Ubuntu 24.10 and Ubuntu 24.04 LTS are the officially supported operating systems for building GrapheneOS

Options are:
- add instruction for each (Idk if GrapheneOS project OK with that)
- add link on documentation on enabling and configuring zswap and zram for each (Idk if GrapheneOS project OK with that)
- add note that user should refer and read instruction on that for used distribution

I would like to get answer on that from GrapheneOS side before making changes on this

static/build.html
flexxxxer: Done as described in my previous comment

static/build.html
thestinger: The recommendation should be to use zswap for the vast majority of developers who have an SSD. It only makes sense to recommend zram for people who an unusual setup such as still using an HDD with a decent amount of RAM such as 32GB. It may still make sense to use zswap instead of zram with 16GB RAM since the point of having actual swap is mainly to avoid losing page cache that's being heavily used and will need to be read back in from storage again. It's possible zram is better with an HDD when tuned well for it but it's quite possible zswap is still better for this use case. zram is more for heavily trying to avoid using storage due to having a low quality SSD without good write durability, latency concerns (but losing page cache can be even worse) and similar reasons. Android uses zram with very conservative use of an idle page writeback file but it might be better off with zswap and a swap file in 2026 due to having decent quality SSDs in the devices. It might be worth us investigating it ourselves for GrapheneOS.

For building GrapheneOS, zram doesn't usually make sense even with a decent SATA SSD and definitely doesn't with an NVMe SSD.  Some distributions including Arch have zswap enabled by default with good defaults so the only configuration required is having either a swap file or swap partition.

Swap files are generally easier since they can be added after the fact and replaced with a different size at any time. The only downside of a swap file is if you want to have per-boot encryption keys for the swap file and don't have a partition without full encryption to put it to avoid the overhead of double encryption. In that case a swap partition is useful to have a different encryption approach for it.

zswap fits into the overall Linux memory management system very nicely and doesn't require manual tuning. There are few knobs to tune since it dynamically adjusts and fits in very well with how memory is managed. zram is a completely different story.

Strongly recommend reading this post:

https://chrisdown.name/2026/03/24/zswap-vs-zram-when-to-use-what.html

flexxxxer: I’ve read this article before and I’ve read it well. still, some users may not or cannot or prefer to not have swap partition/swap file. In PR changes I have noted:
> consider using zswap (require swap partition, preferred if you have NVME storage) or zram (does not require swap partition)

On the main review proposal: what about instructions to setup zswap/zram in GrapheneOS building guide? @ryrona2 propose to add commands in building guide for copy-pasting in terminal in order to setup zswap/zram. Should we add them at all or link to distro-specific sources on how to setup that or add note like "you should refer to instruction on zswap/zram setup for used distribution"?

ryrona2: Are the commands really different per distro? I thought at low level all use modprobe and zramctl and such, even if some offer different ways to do it too at higher level. But I mean, just getting zram mentioned at all was enough for me to get the build running, since I hadn't even thought about the idea, so if it is too tricky to include exact commands, just skip it.

flexxxxer: Ok. Resolved then by not adding instructions on zswap/zram setup

ryrona2 (approved):
Looks good to me. I guess a GrapheneOS developer has to approve the changes too before submitting.

PR #1607, opened 2026-08-04
updated docs to include what iuser device state should b

0xopahc:
I hit the toggle to unlock bootloader in settings -> enter boot menu couldnt unlock bootloadder

I know it seems easy and users should know... but it took me 3x going into settings and hitting the toggle untill it actually took the change on restart

woulda saved me time if docs mentioned what ideal device state shouldbe .. Maybe my wording could be diffrent & maybe belings in diffrent place in the docs??  lmk yall's thoughts on this addition thanks for maintaining good shit

### DIFF
--- a/static/install/web.html
+++ b/static/install/web.html
@@ -235,6 +235,11 @@
                 and the words "Fastboot Mode".  You must not press the device's power button
                 to activate the "Start" menu item, because the device must remain paused in
                 Fastboot mode for the installer to connect to it.</p>
+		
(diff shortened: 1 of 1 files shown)


thestinger:
We could document this but it needs to be done differently. It would be better to file an issue.

0xopahc:
alright thank you  for quick feedback@thestinger 
in a heard of sheep in this world your one of the  few Goats . ty for building this fr 
https://github.com/GrapheneOS/grapheneos.org/issues/1608

Issue #1608, opened 2026-08-04
document what device state should be in bootloader


0xopahc:
something like this would be nice so users not firmilar with bootloader get a 'red/green ' if device state === locked you did last step wrong 
&& if deviceState === locked(unlockable) you did last step correctly 

where should this be documented on static/install/web.html?? 

## closed  Pr reference to understand the idea more..
https://github.com/GrapheneOS/grapheneos.org/pull/1607
lmk what yall think thanks for building good shit

PR #1610, opened 2026-08-05
Update.web install docs

0xopahc:
this looks better then my last one and is on both cli and web .html files
would like feedback on if this fits where it is  now under enableing oem unlocking..

or if it would fit better under Booting into the bootloader interface
and any other comments on adding this to docs 

thanks for building this everyone

####ISSUE 
https://github.com/GrapheneOS/grapheneos.org/issues/1608
#### last pr
https://github.com/GrapheneOS/grapheneos.org/pull/1607

### DIFF
--- a/static/install/cli.html
+++ b/static/install/cli.html
@@ -188,7 +188,7 @@
                 the device to fix OEM unlocking.</p>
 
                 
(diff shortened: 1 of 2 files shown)


0xopahc:
done updating till i get feedback

PR #1635, opened 2026-08-29
[FAQ] Pixel 11 does not meet requirements

qikpl4:


thestinger:
It needs to be more specific than that because the 6th/7th generation Pixels don't have MTE, PAC or BTI.

qikpl4:
> It needs to be more specific than that because the 6th/7th generation Pixels don't have MTE, PAC or BTI.

Fixed.

PR #1655, opened 2026-09-18
Remove BTC BIP47 + PayNym donations

julesh85:
- Removes that part in Donate page
- Removes BTC BIP47 QR code image


thestinger:
I don't want to do this before we have a replacement for it.

